PR #3, opened 2018-12-30
Fix highlighting and reduce page weight a bit

robinwhittleton:
This fixes (for me) page highlighting in the same way as https://github.com/standardebooks/manual/pull/6 . I think we can also drop jQuery to save 30Kb over-the-wire transfer.


acabal:
Thanks! A few notes:

- `var` should probably be `let` in the JS
- Can you undo README changes; I'll fix the typo in a different commit, and the default PHP testing server is not particularly useful for this situation because the ebook/author scripts rely on URL rewrites which I don't think the PHP web server can actually do.

Eventually once the revised manual is complete we can drop the JS entirely and built HTML from the RST source.

robinwhittleton:
Removed the README changes. Going from `var` to `let` locks us to IE11+; if you’re happy with that I’ll do it but I don’t think it’s an issue here given that we’re scoped to the parent function with a single purpose anyway.

acabal:
Yeah I don't mind. One of the nice things about hobby projects is that we can say things like "screw IE11" :)

PR #13, opened 2019-03-11
Reduce the specific spelling updates from 3 to 2 commands

robinwhittleton:
As these only update existing files, we can use the `-a` switch for git commit to add all changed files simultaneously to the commit. This also fixes the formatting to put each command on its own line.


acabal:
Thanks! At some point I wonder if we should merge this list of stuff into modernize-spelling. But it might be tedious to go through a huge list of search and replace all at once without having a chance to stop.

PR #14, opened 2019-03-15
Ellipses in pagination nav should only appear when there are pages between the current page and the first or last page

mgbennet:
Currently, an ellipses appears in the pagination nav at the bottom of the page before and after the currently highlighted page, to indicate that the list is truncated. It looks like this right now:
SE_css_bad

However, if the user is on page 2, the ellipses is still present, despite no pages existing between the current page and the first. My PR will fix this.
SE_css_nav

It will also work when there are only two pages in the current list of pages.
SE_css_nav2

This is an exceedingly minor nitpick but its been bothering me, so I thought I'd try my hand at fixing it.


acabal:
Excellent, thanks!

Issue #8, opened 2019-01-27
Add a script to easily generate the required ebook structure from a corpus of repos


robinwhittleton:
I guess you’ve got some Github API integration to rebuild the library for every release, but without it I’m finding it difficult (even with the README documentation on expected formats) to build the expected heirarchy of ebooks and ebook data. Would it be possible to add a script that, given a folder containing a set of cloned SE repos could build the expected tree of data and potentially copy it into place?

acabal:
I'll email you the script I use, it is not yet ready for public consumption :)

PyroLagus:
Could I have the script too, please? I've managed to get a server set up, but I'm lost on how to set up the ebooks directory. I've tried to place one of the repos in www/ebooks with the built epubs in the dist folder, but 127.0.0.1:8080/ebooks/ just displays the 404 page.

acabal:
If `127.0.0.1:8080/ebooks/` is a 404, are you certain it's actually hitting `127.0.0.1:8080/ebooks/index.php`? Perhaps your server is misconfigured.

You can set up an example ebook in the website like so: 

```bash
mkdir -p /path/to/se/root/www/ebooks/g-k-chesterton/the-man-who-was-thursday/dist
cd /path/to/se/root/www/ebooks/g-k-chesterton/the-man-who-was-thursday/
git clone https://github.com/standardebooks/g-k-chesterton_the-man-who-was-thursday
```

I think that should be it.

PyroLagus:
127.0.0.1:8080/ebooks/index.php is also a 404, even with the directory tree set up like that. It definitely executes ebooks.php, but I assume it triggers [this exception.](https://github.com/standardebooks/web/blob/master/www/ebooks/index.php#L78)

[This](https://gist.github.com/PyroLagus/bfe4b9a3be9ee9ecd1146cae3e72bfd6) is what I have right now. The server starts up, php files get served correctly, but the ebook listing doesn't work. I'm probably doing something wrong, but I can't see what that would be.

acabal:
Unfortunately I'm not familiar with Vagrant. I suggest removing the catch block in index.php to see exactly what the exception is that's happening. However if you follow the bash script I included above, I don't see why it wouldn't work. (I haven't tested it though.)

PyroLagus:
Removing the try/catch gives me this error:

`
[...]

PR #21, opened 2019-05-27
Add script to download books from https://github.com/standardebooks/ …

PyroLagus:
…and keep them up to date.

Adds the last part needed for #8.


acabal:
Looks good, thanks for the hard work! Sorry for the delay.

Later on I'll adjust this a little to match the existing script style. Things like using the `die()` macro, long option names, etc. For now that's not too important.

PyroLagus:
No problem. GNU getopt supports long options, but that would only work on Linux systems, so I just stuck with using short options. It might be possible to manually use shift to go through the args, but I'm not sure that would be done atm.

acabal:
`deploy-ebook-to-www` handles long options. It's rough but effective, especially considering that the use cases here are very simple.

PR #22, opened 2019-06-07
Update Vagrant files.

PyroLagus:
Update Vagrant files to use new composer workflow, automatically install se-tools, and fix xml serving.


acabal:
Thanks!

PR #23, opened 2019-06-07
Fix, clean up, and extend scripts.

PyroLagus:
Mostly clean up, a bit more error handling, some bugfixes, and some extra cli parameters for improved flexibility.

I'm not sure, however, if the --weburl parameter in sync-ebooks is really useful, because even setting that and changing the SITE_URL constant isn't going to make the site, or even just opds, work with a custom URL, since the standardebooks URL is still hardcoded in a few places, such as `web/www/opds/index.xml`. I don't think it hurts to have that option in there, but since the commit is the last one, it's easy to drop if preferred.


scripts/deploy-ebook-to-www
```diff
@@ -99,10 +153,13 @@ do
 	convert -resize "196" -sampling-factor 4:2:0 -strip -quality 75 -colorspace RGB -interlace JPEG "${imgWorkDir}/${urlSafeIdentifier}.png" "${imgWorkDir}/${urlSafeIdentifier}.jpg"
 	convert -resize "392" -sampling-factor 4:2:0 -strip -quality 75 -colorspace RGB -interlace JPEG "${imgWorkDir}/${urlSafeIdentifier}@2x.png" "${imgWorkDir}/${urlSafeIdentifier}@2x.jpg"
 
-	mv "${imgWorkDir}/${urlSafeIdentifier}@2x.jpg" "${imgWorkDir}/${urlSafeIdentifier}-cover@2x.jpg"
-	mv "${imgWorkDir}/${urlSafeIdentifier}.jpg" "${imgWorkDir}/${urlSafeIdentifier}-cover.jpg"
+	hash="$(git log -n 1 --pretty=format:"%ct %s" | sha1sum | awk '{ print substr ($1,1,8) }')"
```
acabal: In this block, it's not necessary to do anything different from the original. The web server accepts requests for the hero that include a unique hash, so that the images aren't cached if they're changed. Apache then rewrites those requests to remove the hash and serve the image generated by this script instead.

scripts/generate-opds.php
```diff
@@ -1,19 +1,24 @@
 <?
-$contentFiles = explode("\n", trim(shell_exec('find /standardebooks.org/www/ebooks/ -name "content.opf" | sort') ?? ''));
+$longopts = array("webroot:", "weburl:");
+$options = getopt("", $longopts);
+$webRoot = $options["webroot"] ?? "/standardebooks.org";
+$webUrl = $options["weburl"] ?? "https://standardebooks.org";
+
+$contentFiles = explode("\n", trim(shell_exec('find ' . $webRoot . '/www/ebooks/ -name "content.opf" | sort') ?? ''));
```
acabal: Make sure to use escapeshellarg() on any user input passed to shell_exec(). In this case we would do `escapeshellarg($webRoot . '/www/ebooks/')`

scripts/generate-rss.php
```diff
@@ -1,6 +1,11 @@
 <?
+$longopts = array("webroot:", "weburl:");
+$options = getopt("", $longopts);
+$webRoot = $options["webroot"] ?? "/standardebooks.org";
+$webUrl = $options["weburl"] ?? "https://standardebooks.org";
+
 $rssLength = 30;
-$contentFiles = explode("\n", trim(shell_exec('find /standardebooks.org/www/ebooks/ -name "content.opf" | sort') ?? ''));
+$contentFiles = explode("\n", trim(shell_exec('find ' . $webRoot . '/www/ebooks/ -name "content.opf" | sort') ?? ''));
```
acabal: Again, we have to use escapeshellarg()

scripts/generate-rss.php
```diff
@@ -65,7 +71,7 @@
 			<pubDate><?= $published ?></pubDate>
 			<guid><?= $url ?></guid>
 			<? foreach($seSubjects as $seSubject){ ?>
-			<category domain="standardebooks.org"><?= htmlspecialchars($seSubject, ENT_QUOTES, 'UTF-8') ?></category>
+			<category domain="<?= preg_replace('/^[^:]+:\/\//ius', '', $webUrl) ?? '' ?>"><?= htmlspecialchars($seSubject, ENT_QUOTES, 'UTF-8') ?></category>
```
acabal: I don't think we need to customize this line. The domain of the category is always going to be standardebooks.org because that's the name of the project.

scripts/sync-ebooks
```diff
@@ -119,29 +156,25 @@ if [ "${verbosity}" -gt 0 ]; then
 	printf " Done.\n"
 fi
 
-urls=$(printf "%s\n" "${urls}" | grep -v -e "/tools.git\$" -e "/web.git\$" -e "/manual.git\$" | awk 'NF')
+repourls=$(printf "%s" "${repourls}" | grep -v -e "/tools.git\$" -e "/web.git\$" -e "/manual.git\$" | awk 'NF')
```
acabal: Please camelCase `repourls` for consistency: `repoUrls`

acabal:
Hi, I've added a few comments for you to review.

How well have you tested these scripts? `deploy-ebook-to-www` and its helpers are running live in production and have been well tested. If something breaks in the deploy process it will be a pain to debug in real time and we'll be left with a broken website while we fix things. I want to make really sure that you've carefully tested all the various permutations of this script to make sure there's no breakage.

scripts/deploy-ebook-to-www
PyroLagus: Oh, I see. That means I'll have to change the Vagrant setup instead.

PyroLagus:
I've tested them several times, but I'll test them again after I fix up the parts you commented.

PyroLagus:
I tested this by cloning all ebooks and using deploy-ebook-to-www in a Vagrant VM. Everything seems to be working correctly.

acabal:
Excellent, thanks!

PR #24, opened 2019-06-13
Add ability to sort by reading ease and length

tuskpot:
As discussed on the mailing list, this set of changes adds two sorting options to the ebooks list page: sort by reading ease, and sort by length (word count).


acabal:
Great, thanks!

PR #27, opened 2019-06-20
Fix vagrant provision script.

PyroLagus:
Some ebooks require firefox for building, and image cover urls as well as new-releases.xml need a rewrite. Perhaps those rewrites should be mentioned in the readme, too.

### DIFF
--- a/scripts/vagrant/provision
+++ b/scripts/vagrant/provision
@@ -12,7 +12,8 @@ set -e
 set -o pipefail
 
 # Install some pre-flight dependencies.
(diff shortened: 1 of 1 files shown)


acabal:
Firefox is only required for the `se compare-versions` command. I decided to leave it out of the regular requirements because it's very heavy weight, for an infrequently used tool. If someone needs that tool then they can make the choice to install Firefox at that time.

PyroLagus:
At least one book wanted firefox for some MathML. I think it was https://github.com/standardebooks/ludwig-wittgenstein_tractatus-logico-philosophicus_c-k-ogden

acabal:
Ah yes, forgot about that!

PR #28, opened 2019-06-21
Add --update-only option back to sync-ebooks.

PyroLagus:
I forgot to add the --update-only option back in during refactor.


acabal:
Thanks!

PR #39, opened 2019-09-17
Update the production guide to use pipx

michael-77:
We are now using pipx rather than pip3 so the step-by-step guide should reflect this.


acabal:
Thanks!

PR #41, opened 2019-10-08
Add instructions for enabling necessary Apache modules in README.md a…

PyroLagus:
…nd add note for vagrant up command


acabal:
Thanks!

PR #44, opened 2019-10-15
Dark mode support

robinwhittleton:
As we support dark mode in our books I thought it’d be nice to add it to the site. This [works in modern browsers](https://caniuse.com/#feat=prefers-color-scheme) (tied to the OS settings). My solution falls back to standard light colours for older browsers: it’s not a perfect version but it’s workable and they’re disappearing fast. Screenshots:

Top of homepage
Center of homepage
Ebook library page
Example ebook page

### DIFF
--- a/www/css/core.css
+++ b/www/css/core.css
@@ -93,13 +93,14 @@
 }
 
 html{
+	color: #222; /* fallback for IE */
 	color: var(--body-text);
 	font-family: "Crimson Pro", Georgia, serif;
 	-webkit-hyphens: auto;
 	hyphens: auto;
 	font-size: 22px;
 	line-height: 1.5;
-	background: url('/images/paper.png') #fcf5dd; /* #faf5e3; */
+	background: url('/images/paper.png') #fcf5dd;
 }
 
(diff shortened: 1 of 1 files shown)


acabal:
Awesome! I'm going to give this a closer review next week but I'm a big fan of dark mode in everything.

acabal:
When you mention older browsers do you basically just mean IE, because it doesn't support CSS vars?

It looks like dark mode prefs aren't supported by Android Chrome, but since it supports CSS vars we should have no problem. It would just display as light mode. Is that correct?

robinwhittleton:
Sorry, should have been clearer about that. Yes, the system defaults to light mode for browsers that don’t support `prefers-color-scheme` as a media query. The fallbacks are just there for browsers without CSS variable support; basically IE at this point.

acabal:
OK awesome, great work!

I'm going to merge it and I want to make some minor tweaks, mainly to increase contrast to the body text. I also want to play around with the darkness of the background so that maybe we can get some contrast against the header. Hopefully I'll have it live in a few days.

PR #45, opened 2019-10-16
Reduce the size of the ebook heading and padding on mobile

robinwhittleton:
A lot of our ebook titles are way too wide on mobile (up to 500px). This reduces the size of the text a bit, but also reduces the padding on either side to give it more room.

| Before | After |
|:-------|:------|
| Screenshot 2019-10-16 at 17 05 04 | Screenshot 2019-10-16 at 17 05 17 |
| Screenshot 2019-10-16 at 17 05 55 | Screenshot 2019-10-16 at 17 06 10 |


acabal:
Great work, thanks!

PR #46, opened 2019-10-26
Update wanted ebooks list

tuskpot:
This update removes two books from the wanted list that have been completed and published.


acabal:
Thanks!

PR #49, opened 2019-12-01
Fix search field styling in Safari

robinwhittleton:
Not sure when this broke, but this disables the default search field style that has shown up. On linght mode it’s ugly but usable, but in dark mode it’s impossible to read your search terms.

Before:
Screenshot 2019-12-01 at 13 14 54

After:
Screenshot 2019-12-01 at 13 15 23


acabal:
Weird, thanks!

PR #52, opened 2019-12-21
Refine #long-description with "authority", "meta-auth" deprecated

jggimi:
When I use "meta-auth" I receive the following warning with 

$ se build -c -o ../dist .

epubcheck v4.2.2 failed with:
WARNING(RSC-017): /home/josh/se/dist/poul-anderson_short-fiction.epub/epub/content.opf(94,58): Warning while parsing file: Use of the meta-auth property is deprecated

Check finished with warnings
$


acabal:
Thanks, that property was removed in epub 3.2 and the meaning of "authority" is not the same. Instead we simply removed that line entirely. Check the template content.opf file in the tools for an example. Can you just remove that line? Thanks!

jggimi:
Sure, will revise the pull to remove the directive from the manual.  Will
double-check my previously published Andre Norton book too and if it has
the line, I'll submit a pull request to correct.

acabal:
Thanks!

PR #55, opened 2020-02-19
Update wanted ebooks list

tuskpot:
This adds some novels by James Stephens to the wanted ebooks list, as discussed recently on the mailing list.


acabal:
Thanks!

PR #56, opened 2020-02-21
List individual Wodehouse novels in wanted ebooks list

tuskpot:
List each Wodehouse novel on the wanted ebooks list, so that new producers can pick one more easily.


acabal:
Thanks!

PR #57, opened 2020-03-26
Wanted ebooks: remove one that has been produced already

jupdike:
and add one first production suggestion that is around 67,000 words


acabal:
Thanks!

PR #63, opened 2020-06-01
Added Burroughs and Verne

b-t-k:


acabal:
Awesome, thanks for taking care of this!

PR #70, opened 2020-07-01
Replace Source Code Pro with Fira Mono for code snippets

robinwhittleton:
The main reason to do this is to fix https://github.com/standardebooks/manual/issues/9, but the font is also about 20% smaller which is a nice benefit.

Fira Mono is also licensed under OFL so no change there. I have no stylistic preference either way, but they’re broadly similar in look. Fira Mono has no italic faces, but we have no italic display of source code so no loss there.

Comparison images:

| Source Code Pro  | Fira Mono |
|---|---|
| Source Code Pro displaying HTML | Fira Mono displaying HTML |
| Source Code Pro displaying terminal output | Fira Mono displaying terminal output |


acabal:
Thanks!

PR #71, opened 2020-07-18
Update wanted list

jupdike:
Added Saki (H. H. Munro). Most of his collections should be beginner friendly productions. It is quite likely an omnibus Short Fiction may be called for, to collect the rest of his short stories, or any non-overlapping collections that might be too short to be published on their own (Reginald, Reginald in Russia). Lots of posthumous material too.


acabal:
If we have to do research on the corpus then that's probably a moderate-difficulty task. We want people to be familiar with the collections policy as well as the production process, before we start exploring an author's body of work when it's not obvious how to approach it.

jupdike:
How do you feel if I add Saki short stories and novels to moderate difficulty section and submit a PR?

acabal:
Sure!

jupdike:
Made another PR for Saki

PR #72, opened 2020-07-26
Better background colour for homepage splash

robinwhittleton:
If we set a dark background colour as well as the image it’ll show while the image is loading (or if the image fails to load). This colour is picked from the background of the image to reduce impact when it loads.

Here’s a screenshot with the main image purposefully toggled off.

Demo without image


acabal:
Good idea, thanks!

PR #73, opened 2020-07-27
Break out novels from short stories on wanted list

tuskpot:
As requested in the Fritz Leiber Short Stories email chain, this moves any "short fiction" entries from "first production" section to "moderate". It also adds an entry for each specific novel (40K words or longer) that is available on Project Gutenberg for those authors.

Note that Fritz Leiber doesn't have any novels on PG, only short stories and novellas.


acabal:
Excellent, thanks!

PR #74, opened 2020-07-29
Add --bare option, default to working directory clones

vr8hub:
It wasn't as easy as just removing the `--bare` from the `git clone` command. The script has a lot of logic revolving around the repository name, and that name is different for a bare (x.git) vs working directory (x) clone.

This makes the default working directory clones, and adds a `-b/--bare` option to get a server-side copy. I made WD the default because I guessed more people use that than use the server-side, plus I think it's the "least surprise" option. And someone needing a service-side copy know they need --bare.

I have tested it both with and without `--bare`, and everything seems to work. I get the proper kinds of repositories based on the option, and both updating and getting new ones work with and with the --bare. I'm currently downloading a complete copy of WDs; it's been going a half-hour or so without incident.

A couple of notes:
1. The `usage()` function at the top used `fmt` to output the text. That reformats everything as a paragraph that wraps at 65 characters. This meant each of the "With…" lines did not start on a separate line, but instead that whole block was treated as a single paragraph. That in turn caused options to get "lost" in the middle of a line. I tried two solutions:
    a. Putting a blank line between each of the "With…" lines, so that each is treated as a separate paragraph.
    b. Changing the `fmt` to `cat`.
The former still wraps everything at 65 characters, but adds an unnecessary empty line in between each option.
The latter doesn't wrap, but I don't see a big need to do that, anyway. Use whatever terminal width you have.
If you prefer the wrapping, just change it back to `fmt` and add an empty line after each "With…". (I really don't think you should leave it like it is; you can't "see" the options.)
2. I use different variable names in a couple of places, but that's because the existing script re-does work on a variable multiple times instead of just using a different variable where it's already done the work.
3.  I added a few comments to make it a little easier to tell what was happening and why.
4. I know you didn't write it, but do you know what lines 186-188 (checking repoName is >= 100 chars, etc.) are doing? Or, more to the point, _why_ it's doing it?
5. There's logic at the end of the script to rename the directory if the SE identifier is different than the repository name. Why would those ever be different?
6. I didn't try to deal with the errors I got last time at the very end of the clone. I'll try to look at that later.

### DIFF
--- a/scripts/sync-ebooks
+++ b/scripts/sync-ebooks
@@ -3,7 +3,7 @@ set -e
 set -o pipefail
 
 usage(){
-	fmt <<EOF
+	cat <<EOF
 DESCRIPTION
 	Syncs books from standardebooks.org GitHub org to specified folder.
 
@@ -13,6 +13,7 @@ USAGE
 	With -v or --verbosity 1, display general progress updates.
 	With -vv or --verbosity 2, display general progress updates and verbose git output.
 	With --update-only, only sync existing repositories, do not download new repositories.
+	With -b or --bare, clone a bare repository (for a server) instead of a working directory
 	With --token TOKEN, specify a GitHub access token to use for request. Useful for when you hit the rate limit.
 
 	DIRECTORY should be where the repositories should go.
@@ -48,6 +49,7 @@ verbosity=0
 updateOnly="false"
 githubToken=""
 target=""
+bare=""
 
 while [ $# -gt 0 ]; do
 	case "$1" in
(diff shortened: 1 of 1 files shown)


vr8hub:
Turns out the error was pretty simple, so I committed another change.

Since there are a mixture of ebook and non-ebook repositories all under the same main SE URL, the script has a line (171) that removes the non-ebook URLs from the repo list. That list had not been updated since the Sublime Test plugin repository was created. I've added it and that fixed the error and a subsequent run downloaded another 30 or 40 repositories after that and completed without error.

Something to keep in mind in the future; if a new non-ebook repository is added, the script should be modified to add it to that line. Or some other way to identify non-ebook repositories?

acabal:
Great work, thanks!

PR #75, opened 2020-07-29
Remove already produced ebook, add Dunsany novel

jupdike:
Removed Verne, Journey to Center of the Earth because it has been produced and released. Added Dunsany's The King of Elfland's Daughter, just added to PG in January 2020, first published in 1924. It is under 70,000 words.


acabal:
Thanks!

PR #77, opened 2020-08-01
Add check to prevent script dying

vr8hub:
I added a few more comments to better document things, including a note at the top that it requires the GNU versions of grep/sed, so that Mac users wouldn't be caught unawares.

Functionally, I added a check to make sure that a repository has a metadata file before it does the part that will rename the repository if it doesn't match the SE identifier name. This prevents the error I was getting even if a non-ebook repository slips by the check in the script. (Like would happen if a new non-ebook one is added and the script isn't updated.)

I tested it by commenting out the line (197 in this version) that removes the non-ebook repositories and then running a sync. It dutifully added all of the non-ebook repositories, i.e. tools, web, etc., without erroring.

The script is still filtering them out, but if one slips by that isn't in the list, it will no longer error and thereby cause someone to miss the remaining ebook repositories that follow that non-ebook one.

### DIFF
--- a/scripts/sync-ebooks
+++ b/scripts/sync-ebooks
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env bash
 set -e
 set -o pipefail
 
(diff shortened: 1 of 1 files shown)


vr8hub:
Because the `ls` used for the update existing repository section was unrestricted, if there were any files present in the directory, the script would crash. I added a small change to only look for directories, so any files would be ignored. No, there probably shouldn't any files in the directory, but if there are, and someone starts a complete sync and goes to bed, I'd hate for them to get up the next morning and find nothing happened. :)

vr8hub:
And also prevent a crash if a non-git directory manages to find its way in.

acabal:
Great work, thanks!

PR #79, opened 2020-08-20
Fix typo: Add missing word 'has'

maticstric:
Don't know if this is exactly how we want to fix it, but it looks like there's a missing word in: "Fortunately, Standard Ebooks tools for that too"


acabal:
Thanks!

PR #85, opened 2020-10-29
Add a guide for Shakespeare's play

EmmaSweeney:


acabal:
Great Emma, thanks. Let's structure it so that there's an index page in How Tos that leads to subpages, and Sheakespeare can be the first (and for now only) one. That way when we add more how-tos they reader doesn't get overwhelmed by one giant page. Can you do that?

EmmaSweeney:
I think I'm gonna need some help with this. My understanding of PHP files is very limited.

acabal:
First create a new folder, `how-tos`. Then in that folder create index.php, and fill it just like this file, but all it will contain (for now) is a link to the Shakespeare how-to. Move the file in this pull request file into `hos-tos` and call it `shakespeare.php`. Its URL will be `/contribute/how-tos/shakespeare`.

EmmaSweeney:
OK, that makes sense. I'll go work on that now. Thanks!

acabal:
Excellent work, thanks!

PR #88, opened 2020-11-22
Remove duplicate Boswell from wanted list

irontigran:
We had two entries on the wanted list for Boswell's Life of Johnson.
Also clear up which edition and volume the PG link is pointing to.

### DIFF
--- a/www/contribute/wanted-ebooks.php
+++ b/www/contribute/wanted-ebooks.php
@@ -560,7 +560,7 @@ require_once('Core.php');
 				<p><a href="https://books.google.com/books?id=OKREAAAAYAAJ">Cyrano de Bergerac</a>, by Edmond Rostand, translated by Brian Hooker. This is a 1923 publication, transcription is required.</p>
 			</li>
 			<li>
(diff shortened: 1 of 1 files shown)


acabal:
Thanks! Are you certain that edition is the unabridged edition? How can you tell?

irontigran:
Mostly going off the Wikipedia article on [Life of Johnson](https://en.wikipedia.org/wiki/Life_of_Samuel_Johnson#Notable_editions)

To paraphrase: three 'initial' editions were published in the 18th century, all of which were unabridged. George Birkbeck Hill's edition (published in 1887, in six volumes) uses the third edition as its basis, then adds its own commentary on top of the existing notes.

[Gutenberg](https://www.gutenberg.org/ebooks/author/659) has has two copies of _Life of Johnson_; one is clearly labeled abridged and edited by Charles Osgood. The other is split into six volumes, each marked as edited by Hill. Without claiming any particular expertise, I feel comfortable saying that those six are Hill's six volumes.

acabal:
OK, that makes sense. Probably when I looked at it last (a long time ago) the abridged version may have been the only option at PG, or maybe I just overlooked the other ones. Thanks!

PR #89, opened 2020-12-07
Reduce size of homepage SVGs

guilhermesimoes:
In the spirit of [SuperTinyIcons](https://github.com/edent/SuperTinyIcons), I decided to see if I could reduce the size of the SVG files used in this project's homepage.

The homepage SVGs went from 76kB to 29kB.
The meta logo went from 29kB to 5kB.

I also reduced the size of the logo-full (which is not used in the homepage) from 34kB to 10kB because it's exactly the same as the logo-white-full (which is used in the homepage) except its color.

You can use [GitHub's rich diff](https://github.blog/2014-10-06-svg-viewing-diffing/) to see that the SVG files remain the same visually.


acabal:
Great work, thanks! Did you do this by hand or with a program?

What I'll probably do is remove logo-white-full on the homepage, and just use logo-full with an invert filter to make it white.

guilhermesimoes:
One part was by hand. One big gain was in moving half of the laurel wreath into a group, and then [using it on both the left and right sides, while flipping it on the right side](https://github.com/standardebooks/web/blob/5c4617d759a2a15773ea39a05e8d146bcd8872ea/www/images/logo.svg?short_path=cac6f1c#L37-L38).
But then I used [SVGOMG](https://jakearchibald.github.io/svgomg/) to remove all the unnecessary metadata added by the programs that created these SVG files.

> What I'll probably do is remove logo-white-full on the homepage, and just use logo-full with an invert filter to make it white.

Instead of doing that, can I make a suggestion? It would require you to embed the SVG directly in the page instead of using a background image but it would give us more control over the SVG color and make it unnecessary to use a filter.

Basically, instead of doing this:
```html
<a href="/">Standard Ebooks</a>
// with a ::before with a background-image: url(...logo.svg)
```
if we did this:
```html
<a href="/" aria-label="Standard Ebooks">
  <svg>...</svg>
</a>
```
We could then go to the SVG file and change the fixed `fill` and `stroke` colors:
https://github.com/standardebooks/web/blob/master/www/images/logo-full.svg?short_path=8db01d2#L1
https://github.com/standardebooks/web/blob/master/www/images/logo-full.svg?short_path=8db01d2#L31
from `#222` to [`currentColor`](https://css-tricks.com/cascading-svg-fill-color/).

This way the SVG will inherit any color we give it through CSS :smile:

acabal:
Yeah, but I'd rather not embed the SVG into the page. It prevents the SVG from caching (since the PHP page has caching disabled) and it makes the page source much less manageable. Using filter works just fine and if a user wants to save the image to disk it will save with dark text on a transparent background which is ideal.

guilhermesimoes:
Makes sense!

PR #90, opened 2020-12-07
Fix homepage link preview

guilhermesimoes:
I shared a link to this project with a friend on whatsapp and there was no link preview:

IMG_4932

[According to facebook](https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fstandardebooks.org%2F):

> Unsupported Image File Extension
> Provided og:image URL, https://standardebooks.org/images/logo.svg does not have a supported extension.

I swapped the SVG logo for the homepage image because:

1. It's more visually appealing
2. It's what facebook is already using

I also took a look at twitter and there was no link preview. [According to twitter](https://cards-dev.twitter.com/validator):

> ERROR: Required meta tag missing (twitter:text:title)

[According to twitter's markup](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/markup), `og:title` is also accepted but on this page it's empty.

I added a fallback for this case.

Hopefully these two changes will fix the link previews.

### DIFF
--- a/templates/Header.php
+++ b/templates/Header.php
@@ -42,7 +42,7 @@ print("\n");
 	<link rel="alternate" type="application/rss+xml" title="Standard Ebooks - New Releases" href="https://standardebooks.org/rss/new-releases"/>
 	<link color="#394451" href="/safari-pinned-tab.svg" rel="mask-icon"/>
 	<meta content="#394451" name="theme-color"/>
-	<meta content="<?= Formatter::ToPlainText($title) ?>" property="og:title"/>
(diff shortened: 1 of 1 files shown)


acabal:
Excellent, thanks for sorting this out!

guilhermesimoes:
The preview changed on Slack :smile:

| Before | After |
|---------|--------------|
| Screenshot 2020-12-08 at 21 03 10 | Screenshot 2020-12-08 at 21 03 31 |

Unfortunately it still doesn't seem to work on Twitter or Whatsapp. Maybe their previews are still cached, and they'll be fixed in a few days. Unfortunately I don't know what else to do.

PR #97, opened 2020-12-29
Put the burden of compatibility on devices and not the books

robinwhittleton:
This rewrites the messaging around our downloads to make it clearer what we offer. I think it was clear from the recent MobileRead thread that there’s some misunderstanding with what we offer, and hopefully by moving the compatibility statements from our books to the ereaders we could clear some of them up.

### DIFF
--- a/www/ebooks/ebook.php
+++ b/www/ebooks/ebook.php
@@ -155,7 +155,7 @@ catch(\Exception $ex){
 						<span><a property="schema:contentUrl" href="<?= $ebook->EpubUrl ?>" class="epub">
 						Compatible epub
 						<meta property="schema:encodingFormat" content="application/epub+zip"/>
-						</a></span> <span>—</span> <span>All compatible devices and apps, except Kindles and Kobos.</span></p>
+						</a></span> <span>—</span> <span>All devices and apps except Kindles and Kobos.</span></p>
 					</li>
 					<? } ?>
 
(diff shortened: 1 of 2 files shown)


acabal:
These changes are OK in general, except that I don't want to remove the compatible label, because we want to encourage almost everyone to download that file. As I stated in that thread, "compatible" must have some reasonable definition, and just because it doesn't meet the definition of some guys on a forum using abandonware from 2012, does not make it incompatible. We would not expect a compatible Word document prepared using MS Word 2020 to work on Word 1.0 for MS-DOS from 1987, would we?

*Maybe* the more accurate phrasing would be "compatible epub 3" file. (Because at the core of it, their beef is that we are not shipping epub 2 files.) *But*, that invites confusion from less technical people as to, "does my ereader support epub 3?" The answer is of course almost always yes, except for those few guys on Mobileread. Saying "all compatible devices" invites the same question: Is my device compatible? And the answer is the same as above, so that can also be confusing.

And, as I also stated in that thread, this issue has literally never come up in 6+ years of millions of ebook downloads. It's not a real problem.

robinwhittleton:
I’m totally happy to not take this change if you think it’s solving a non-existent problem 🙂

One thing I would say is that we should probably scrub references to epub2. There’s the one in this change of course, but we’ve got a few comments throughout the toolchain as well.

acabal:
If you want to tweak this commit to remove the "compatible" changes, then go for it,  I think the rest is good. You can also tweak any epub2 comments in the toolchain. Long ago we *did* release epub2 files but that hasn't been the case for a long time now.

robinwhittleton:
Sorry, wasn’t 100% on what you meant. I’ve put “Compaptible” back before epub in the title, and added one more compatible into how-to-use-our-ebooks. Think this this OK? Happy to make more changes.

acabal:
Thanks!

PR #98, opened 2021-01-02
Make font white for book number in collections view

aleyan:
In the collections view, there is a dark square (`rgba(0, 0, 0, .75)`) containing the ordinal number for the book, however because the font is dark, the text isn't visible. This PR sets the font in that selector to white. Not sure if pure `fff` white is the right color here, so looking for feedback on the the right shade/variable to use.

Before:
se_collections_before

After:
se_collections_after


acabal:
Thanks!

PR #100, opened 2021-01-08
Fix wanted ebooks

weijia-cheng:
I think the unescaped ampersand breaks the XML parsing.


weijia-cheng:
Screenshot from 2021-01-08 06-44-20

acabal:
Oops, thanks!

PR #103, opened 2021-01-27
Typo in README.md

thewchan:
I think tstandardebooks.test should be standardebooks.test?


thewchan:
Never mind I understand now; the \t is a tab

PR #108, opened 2021-02-08
Add Docker support

robinwhittleton:
This is specifically designed to provide a suitable environment for developing on a clone of standardebooks.org. It’s not trying to provide a process for containerising SE for deployment.

The instructions provided will set up a SE running in a container, and allow the host to access it. Source code is bound rather than copied. Changes made to the PHP will require a PHP restart inside the container: there’s currently no hot reloading.

I’ll try to keep the Dockerfile updated to match README installation instructions over time. There’s little deviation, apart from the split of “stuff built in the image” vs “stuff run in the container”.

The container doesn’t yet provide support for generating ebooks. If needed, these are expected to exist in the appropriate format already.

(From my specific point of view, this allows me to hack on the CSS on a local copy of the site. I’ve not been easily able to do this in the past.)


Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM ubuntu:20.04
+
+RUN apt-get update
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y composer php-fpm php-cli php-gd php-xml php-apcu php-mbstring php-intl apache2 apache2-utils libfcgi0ldbl task-spooler
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
acabal, line 5: Is this necessary?

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+RUN mkdir /standardebooks.org
+RUN mkdir /standardebooks.org/web
+RUN mkdir /standardebooks.org/web/config
+RUN mkdir /standardebooks.org/web/config/ssl
```
acabal, line 10: This can all be combined into one line with mkdir -p

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN mkdir /var/log/local
+
+RUN openssl req -x509 -nodes -days 99999 -newkey rsa:4096 -subj "/CN=standardebooks.test" -keyout /standardebooks.org/web/config/ssl/standardebooks.test.key -sha256 -out /standardebooks.org/web/config/ssl/standardebooks.test.crt
+
+RUN a2enmod headers
+RUN a2enmod expires
+RUN a2enmod ssl
+RUN a2enmod rewrite
+RUN a2enmod proxy
+RUN a2enmod proxy_fcgi
```
acabal, line 20: These can also all be on one line. a2enmod can enable many modules in one invocation. See the readme

scripts/docker/start-server.sh
```diff
@@ -0,0 +1,13 @@
...
+composer install
+
+ln -s /standardebooks.org/web/config/apache/standardebooks.test.conf /etc/apache2/sites-available/
+ln -s /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/*/cli/conf.d/
+ln -s /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/*/fpm/conf.d/
+ln -s /standardebooks.org/web/config/php/fpm/standardebooks.test.conf /etc/php/*/fpm/pool.d/
+a2ensite standardebooks.test
+/etc/init.d/apache2 start
+service php7.4-fpm restart
+tail -f /var/log/apache2/error.log
```
acabal, line 13: Do we need to do this with the log file? Presumably this docker image works out of the box already

README.md
```diff
@@ -58,6 +58,30 @@ git clone --bare https://github.com/standardebooks/david-lindsay_a-voyage-to-arc
...
+
+Then run the built image with:
+
+```shell
+docker run -dp 443:443 -v "$(pwd):/standardebooks.org/web" standardebooks:latest
+```
+
+The site will now be available at `https://localhost/`, although as it’s a self-signed certificate you’ll need to accept whatever browser security warnings come up.
+
+If you make a change to the PHP and want to see that reflected in the site, then you’ll need to restart PHP. Connect to your container and run:
```
acabal, line 77: I don't think this is correct, generally FPM doesn't need to be restarted for PHP changes to take effect. It *may* need to be restarted if data cached in APCu is changed and the c cache has to be flushed, but that's a different case than just "any PHP code changes."

acabal (changes requested):
I don't know much about Docker, but I wonder why we have both a `Dockerfile` and also a `start-server.sh`. Can't we combine them? It looks all they do is configure the server using several sudo-level commands. This will also make it easier to track changes if the readme install instructions update--I think we want the readme instructions, and any recipes in containers like this, to be as identical as possible.

README.md
acabal: Actually, I think it depends on whether you're installing Ubuntu Desktop or Ubuntu Server. Desktop is configured with the opcache disabled and server is configured with it enabled. So if this specifies the server image then this would be correct after all.

robinwhittleton:
So bear with me, because I’m somewhat of a Docker newbie as well. But to try to answer the non-review question: Docker has the concept of an image, and of a container. The image is the baseline that stays the same over time, and is defined in the Dockerfile. The container is instantiated from an image, and starts at the entrypoint.

In our case, we’re trying to do a development image by mounting the host’s web repo into the docker container. The mounting happens at the point of `docker run`, i.e. after you’ve got a built image and you’re instantiating a container. That means that things that need to be done against the mounted code (i.e. `composer install`) have to be done after the entrypoint. We could move more stuff from the Dockerfile into the bash script, but that would mean more things to be run every time a container is started.

Dockerfile
robinwhittleton: Thanks, easy. https://github.com/standardebooks/web/pull/108/commits/2b4af01b060ead9ba0624717e1460ad145cf5a3b

Dockerfile
robinwhittleton: https://github.com/standardebooks/web/pull/108/commits/2b4af01b060ead9ba0624717e1460ad145cf5a3b

Dockerfile
robinwhittleton: It’s not, but it shrinks the size of the image. The expectation is that you don’t interactively work with the image in the long term. If you’d make a change to the setup you’d do it in the Dockerfile, then rebuild the image again. But I’m happy to remove it if you’d prefer.

README.md
robinwhittleton: Aha, yep, that fixed it. https://github.com/standardebooks/web/pull/108/commits/a66a8c5e853705843ae67725593076a8a172e7dd

acabal:
OK, I suppose that's fine given I don't know anything about Docker. 

Last points, is the `tail` of the logfile necessary; also, since I just removed the test SSL keys from the repo, the ssl dir that we create in the Docker file should already be there as part of the web repo. Do we still need to create it explicitly?

acabal:
One last thought. We already have a Vagrant config which as I understand is basically similar to Docker. I don't want to start littering our repo root with a bunch of config files for a bunch of different virtualization options. Can we group these all together in to a `containers` or `vms` folder or something like that?

scripts/docker/start-server.sh
robinwhittleton: We need to do something, otherwise the container finishes and exits 🙂 A lot of containers never suffer from this problem because they start an undaemonised server (`npm start` or equivalent), but here PHP and Apache both run in the background. A typical hack seems to be `tail -f /dev/null` in the foreground at the end of your script, so I’ve updated my script with that and a comment. https://github.com/standardebooks/web/pull/108/commits/2e28e5c65d6580e25d2aeffc5978eb5ef37f5bde

robinwhittleton:
Regarding the SSL folder, we do seem to need to create it explicitly. I can’t work out quite why, but the image fails if I don’t 🤷‍♂️

robinwhittleton:
Regarding Vagrant, I have to say that it’s been a while since I’ve seen it used, but I’m presumably fairly filter-bubbled. @PyroLagus: are you still using the Vagrantfile? It’s a bit out of date at this point.

If not then I’ll ask on the mailing list. That’ll let me know if it should be retired or maintained.

acabal:
So can we put these in some containers folder or something like that? Is it required that they be in the project root?

robinwhittleton:
OK, I’ve no idea about Vagrant but I’ve created a vms/docker directory and updated the readme instructions. In doing that I found a bug with my code, exposed after you deleted the test keys (😅) so I’ve adjusted that too.

acabal:
I'm looking at the Vagrant files now and it's wildly out of date. And it uses nginx for some reason? I think I'll just get rid of it entirely since nobody is maintaining it and it's broken as of now.

robinwhittleton:
OK! There’s a vagrant folder in the scripts folder too.

vms/docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y composer php-fpm php-cli php-gd php-xml php-apcu php-mbstring php-intl apache2 apache2-utils libfcgi0ldbl task-spooler
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+RUN mkdir -p /standardebooks.org/web
+RUN mkdir /var/log/local
+
+RUN a2enmod headers expires ssl rewrite proxy proxy_fcgi
+
+# Disable opcaching for dynamic PHP reloading
+RUN echo "opcache.enable=0\n" >> /etc/php/7.4/fpm/php.ini
```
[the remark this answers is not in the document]
robinwhittleton: Fixed in https://github.com/standardebooks/web/pull/108/commits/4d7e08f7f2477afcc94ec4d798fb59da5a8d396d.

vms/docker/start-server.sh
```diff
@@ -0,0 +1,19 @@
...
+cd /standardebooks.org/web
+composer install
+
+ln -s /standardebooks.org/web/config/apache/standardebooks.test.conf /etc/apache2/sites-available/
+ln -s /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/*/cli/conf.d/
+ln -s /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/*/fpm/conf.d/
+ln -s /standardebooks.org/web/config/php/fpm/standardebooks.test.conf /etc/php/*/fpm/pool.d/
+a2ensite standardebooks.test
+/etc/init.d/apache2 start
+service php7.4-fpm restart
```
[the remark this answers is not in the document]
robinwhittleton: So this is a super-stripped back container. The Docker idea is that each container effectively does one thing with the minimal resources it needs, and if you need more services (i.e. if we had a database for the site) then you’d spin that up as a separate container and expose the necessary interfaces instead. Indeed, there’s Docker Compose for orchestrating multiple containers for your single app, though I’ve avoided that for this.

Because you’re not running multiple systems on the single container, they seem to have gone with a more stripped back approach:

> System has not been booted with systemd as init system (PID 1). Can't operate.

acabal:
So they made their own image of Ubuntu without systemd? Good God. Not confidence inspiring.

In that case we should use service to start apache as well for consistency

robinwhittleton:
Canonical maintain the Ubuntu Docker image: https://hub.docker.com/_/ubuntu . The page says:

> This image is built from official rootfs tarballs provided by Canonical (specifically, https://partner-images.canonical.com/core/).

I’ll look into the service change.

acabal:
It should just be the same as the fpm invocation. `service apache2 start`. `service` was how this was done in general in Debian/Ubuntu before they switched to systemd. If it's set up for FPM then I can't imagine why it's not set up for Apache.

vms/docker/start-server.sh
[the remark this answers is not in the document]
robinwhittleton: Started Apache in the same way as PHP in https://github.com/standardebooks/web/pull/108/commits/4718db4e8f2feeff806873730892e7668b25537a.

acabal:
OK great. Thanks for bearing with me!

robinwhittleton:
Not at all! All your suggestions made it better.

PR #110, opened 2021-02-13
Improve instructions on dealing with commas and periods within quotes

weijia-cheng:
Provide a helpful interactive-sr command (something I use myself anyways) instead of just telling people the regex. Also a helpful warning.

### DIFF
--- a/www/contribute/producing-an-ebook-step-by-step.php
+++ b/www/contribute/producing-an-ebook-step-by-step.php
@@ -189,7 +189,27 @@ proceed to seal up my confession, I bring the life of that unhappy Henry Jekyll 
 						<p><a href="/manual/latest/8-typography#8.7.7">Two-em dashes should be used for elision</a>.</p>
 					</li>
 					<li>
-						<p>Commas and periods should generally be inside quotation marks, not outside. This regex helps find them: <code class="regex">[’”][,.]</code></p>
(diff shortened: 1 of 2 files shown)


acabal:
I think it might be even clearer if we expanded the example a little bit. What's an example of a plural possessive in a sentence, vs. something that needs correcting?

weijia-cheng:
Ok, how about this? I played around with the manual in a docker container and came up with this. It has examples of what is acceptable and what is not, based on the spacing list. (I used <code> because it's hard to distinguish between the fixed and unfixed versions otherwise.)

acabal:
OK, but it should be `your guys` not `you guys`, and I don't think the comma in `Hello` is grammatically correct either. Maybe just pull a sentence from an existing production as an example instead of trying to come up with our own?

weijia-cheng:
Ok, maybe these should be more compelling examples.

acabal:
`This wood now is not mine, but yours, you peasants’.` is also grammatically incorrect.  You're making `peasants` possessive but you're addressing the plural, not possessive, peasants with `you`.

weijia-cheng:
Ok, I was just copying something I saw in Tolstoy but I can see the problem. I tweaked it a bit to `This wood now is not mine, but the peasants’.` which is more standard grammatically.

acabal:
Great, thanks!

PR #111, opened 2021-02-14
Update wanted books

weijia-cheng:
Removing the commented out _Heretics_, which has been produced now. Also recommend _The Sport of the Gods_ (https://www.gutenberg.org/ebooks/17854) by Dunbar as a possible easy production. I looked at the html and the formatting looks very straightforward at first glance. The book is about 41k words by my count, so a little on the short side but perfect for a first production.


acabal:
Thanks!

PR #112, opened 2021-02-14
Fix formatting of quote fix command

weijia-cheng:
Realized I forgot to add the `<i>` tags to make the string in the command display nicely


acabal:
Thanks!

PR #117, opened 2021-03-19
Condense git commit commands on step-by-step guide

weijia-cheng:
Practically, I don't see any reason why we use the `git add -A` `git commit -m` work flow in certain steps. I think it makes sense to just condense all of these that we can into the one-liner, especially since we're using it already. I have changed every step that (to the best of my knowledge) does not add any new files to `git commit -am`.


robinwhittleton:
Looks good. `-am` will also not remove any deleted files from the repo, but that’s not the case for any of your examples. I’ll leave to Alex to merge.

acabal:
Thanks!

PR #122, opened 2021-03-29
update URL to avoid redirect

partev:


acabal:
Thanks, it looks like Groups changed their URL format since this page was created.

PR #118, opened 2021-03-21
Wanted: Theodore Roosevelt

bentley:
Theodore Roosevelt was a prolific author. I took a look at what’s available from Project Gutenberg. Some notes on the ones I added are below.

I left off his biographies of others, and his political works. His autobiographies seem much more interesting.

First production:

* *The Rough Riders*, 63463 words. A couple of footnotes and letters. One &lt;i> tag, maybe there weren't any to strip?
* *Hunting the Grisly and Other Sketches*, 62626 words. PG preserved the italics.
* *Outdoor Pastimes of an American Hunter*, 122562 words. Lots of photographs, all can be dropped. PG is well‐formatted.
* *A Book‐Lover’s Holidays in the Open*, 80838 words. One song, one dialogue, one letter, and a handful of footnotes. PG is well‐formatted.

Moderate:

* *Theodore Roosevelt: An Autobiography*, 217230 words. Several letters. A handful of footnotes. A couple of lists/tables. PG typography is mixed (contains italics and em dashes but also straight quotes). Marked as intermediate due to length and mild formatting necessary.
* *Through the Brazilian Wilderness*, 112635 words. A couple of lists/letters/tables in the appendices. Marked as intermediate because PG dropped the italics and accents (especially problematic considering the many Portuguese names).
* *The Winning of the West*, four volumes, 418943 words total. Many footnotes, occasional letters, tables, poems, but nothing looking too complicated. PG typography is mixed (contains italics and em dashes but also straight quotes). Marked intermediate due to the length.

Advanced:

* *The Naval War of 1812*, 157904 words. Many tables, lists, etc. Many footnotes. Many illustrations (not present in PG). Most can probably be left out but there are quite a few diagrams (ship movements, etc.) that will need to be reproduced from scans. [Wikipedia](https://en.wikipedia.org/wiki/The_Naval_War_of_1812#Impact): “It is considered a seminal work in its field, and had a massive impact on the formation of the modern American Navy. … It did well, going through four editions in six years. In 1886, just four years after being published, the U.S. Navy ordered a copy of the book to be placed on every ship. … The book also influenced Alfred Thayer Mahan when he wrote The Influence of Sea Power Upon History, considered the greatest work on naval war in history.”

There are several other books scanned but not yet available on PG, including:

* *[Ranch Life and the Hunting-Trail](https://archive.org/details/ranchlifeandhun00remigoog/)* (227 pages)
[...]

PR #127, opened 2021-05-08
Update wanted ebooks: Jefferson, Mahan

bentley:
Revising #118 a bit: it was suggested that we would be interested in Alfred Thayer Mahan’s *The Influence of Sea Power Upon History* over Roosevelt’s *The Naval War of 1812*. I protested that *Sea Power* and its sequel don’t cover the War of 1812. I’ve since learned that Mahan in fact published a third book specifically covering that war, so I agree that Roosevelt’s *1812* is obviated.

I’m removing Roosevelt’s *The Winning of the West*, as it turns out to not be autobiographical, and he is not particularly notable as a historian (except for *1812*).

I looked into other autobiographies of U.S. presidents. I think Jefferson’s is worth adding to the list. It exists on Gutenberg, is about 40k words plus 20k words of appendices, and covers his life up to 1791. I listed it as advanced because it has a few spots with tricky formatting.

Some others that I investigated but decided not to add:

There is a memoir of John Adams within his papers but it is very short. I do think productions of [the Adams–Jefferson letters](https://en.wikipedia.org/wiki/Thomas_Jefferson#Reconciliation_with_Adams) and of John and Abigail Adams’s letters to each other would be worthwhile for SE, but both would require a lot of research.

James Madison’s autobiography exists as a [24‐page manuscript](https://www.loc.gov/item/mjm022931/).

James Buchanan wrote [*Mr. Buchanan’s Administration on the Eve of Rebellion*](https://archive.org/details/DKC0010). It was looked upon disfavorably as an attempt to defend his failure to prevent the Civil War.

Martin Van Buren wrote an autobiography. It is [over 800 pages long](https://archive.org/details/autobiographyma00buregoog) and by all accounts a real yawner.

The diaries of [James K. Polk](https://archive.org/details/diaryofjameskpol00polk) and [James A. Garfield](https://archive.org/details/diaryofjamesagar04harr) during their presidencies have both been published. I must admit I can’t recall anything noteworthy about either President (although in Garfield’s defense, he was shot and died six months in, so he didn’t have much of a chance).

I’m not aware of any other U.S. Presidents who have public domain autobiographies. There may be a few if one digs through collections of their papers but I don’t think it’s worth the effort.

The only one expiring soon is Calvin Coolidge’s (published 1929). That would be a good fit for our collection.

### DIFF
--- a/www/contribute/wanted-ebooks.php
+++ b/www/contribute/wanted-ebooks.php
@@ -734,7 +734,7 @@ require_once('Core.php');
 				<p>Hadrian the Seventh by Frederick Rolfe (No PG edition yet but <a href="https://www.pgdp.net/c/project.php?id=projectID5d6cced48cdf7&amp;detail_level=3">PGDP transcription is in progress</a>. Part of the Guardian 2015 collection.)</p>
 			</li>
 			<li>
-				<p>The Winning of the West (<a href="https://gutenberg.org/ebooks/11941">Volume 1</a>, <a href="https://gutenberg.org/ebooks/11942">Volume 2</a>, <a href="https://gutenberg.org/ebooks/11943">Volume 3</a>, <a href="https://gutenberg.org/ebooks/11944">Volume 4</a>) by Theodore Roosevelt</p>
+				<p><a href="https://gutenberg.org/ebooks/45847">The Autobiography of Thomas Jefferson</a>, by Thomas Jefferson</p>
 			</li>
 		</ul>
 		<h2>Uncategorized lists</h2>


acabal:
The thing is, we generally don't want histories: https://standardebooks.org/contribute/accepted-ebooks

There is an exception for primary sources and very notable histories. _The Influence of Sea Power Upon History_ may be notable because as Wikipedia says "scholars considered it the single most influential book in naval strategy." That sounds notable to me. But are those other histories notable? I don't really think so--they're just histories. That is also why I declined Roosevelt's history book.

Autobiographies are OK. Collections of letters and papers and diaries are probably not unless they can all be categorized as "timeless non fiction" i.e. philosophy, notability, or something like that.

bentley:
I’ve dropped Mahan from the commit, so now it just removes *Winning of the West* and adds Jefferson’s autobiography.

acabal:
Thanks!

PR #128, opened 2021-05-10
Add how-to guide for music figures

ACBSmith:
I've manually written the html to look like the Shakespeare guide, except where I have taken blocks from Pandoc's conversion of the .rst I wrote for tip and code boxes.

I'm unsure about images; as far as I can tell, I'm using the same format that's also on the front page.


ACBSmith:
Hi Alex - any feedback on this?

acabal:
Looks great! A few issues:

- The website is served as XHTML so XHTML rules apply, just like ebooks. `` must be self-closing and there is a missing `</a>` on line 79.
- The manual just got updated, can you update the links to the latest version and confirm that the items are the same? They should be I think.
- Images should be in a `<figure>` element. Currently they are too wide for the screen, creating a horizontal scrollbar. We want them to not extend past the main content area, i.e. they should not go past the text.
- Man and Superman is a book and thus gets italics (but don't worry about semantics, this is not an ebook)
- Is there a Musescore tutorial somewhere out there that we can link to? It can be obscure for someone who's not used to it.
- `are are`
- SVG code examples should go in a `<figure><code>` element, see the manual. Don't worry about syntax highlighting though.
- XML should not be in italics?

I think it would be helpful for you to install the website on your local machine to see if you can preview this in a browser. You can install it in a VM with Ubuntu, or I believe there's a Docker container.

ACBSmith:
Good call to install Docker. That solved a lot of the guesswork.

I think I've addressed all of those issues - I ended up manually swapping out the classes to colour it correctly. I know you said not to worry about syntax highlighting, but it was bothering me. It all looks good to me except the !DOCTYPE line (line 73) - any input there?

I'd like to make the images a little wider, too, but am not sure about if that would mess with house style.

acabal:
First pull from master, then you can add the `full-width` class to the `<figure>` to get the images to be 100% width. You missed coloring the `>` in the XML declaration. Otherwise everything looks good. Let me know when this is wrapped up and I'll merge it in.

ACBSmith:
I think I've merged the master in correctly, but honestly my installation of Sourcetree seems to hate me, so I'm not sure.

acabal:
Looks great, thanks! Excellent work!

PR #129, opened 2021-05-13
Added Goodreads guide and images

drgrigg:
Alex: I fear this is a bit rough around the edges. I wasn't quite sure how to deal with the screenshot images (should they be 100% width, do they need a border, etc), nor where to link this new guide. It doesn't quite fit with the current text on the Guides page of "guides to help produce ebooks that contain complex formatting."

I also can't test the website locally other than by dummying up the content in a html page.


acabal:
Sorry for the delay, I got tied up with the toolset 2.0.0 release.

This looks very good, great work!

drgrigg:
Thanks, but the link from the index goes to the Shakespeare info. Not sure if that was my mistake. Can you fix?

acabal:
Oops, should be fixed now!

PR #135, opened 2021-06-09
Update British to American text to include non-dialog outer quotes

vr8hub:
This came up yesterday/today, but I've seen the comment a couple of other times on the list: because it says "dialog," some people take that literally (and exclusively). This is an attempt to indicate it includes all outer quotes, not just dialog.


acabal:
Thanks!

Issue #136, opened 2021-06-12
sed --in-place isn’t posix compatible


robinwhittleton:
https://github.com/standardebooks/web/blob/241242749097fdecd64a9164f050ba689e25030c/www/contribute/producing-an-ebook-step-by-step.php#L118

Unfortunately switching from perl to sed has broken macOS compatibility, as `--in-place` is a GNU thing. Short of requiring mac users to `brew install gnu-sed` there don’t seem many clean options (see https://stackoverflow.com/questions/5694228/sed-in-place-flag-that-works-both-on-mac-bsd-and-linux).

What was the rationale for switching in the first place?

acabal:
For consistency, since `sed --in-place` was used elsewhere in the guide but perl wasn't. But it looks like most of those have been replaced with `interactive-replace` so if you want to replace all of the instances of `sed` with `perl` then go for it.

PR #137, opened 2021-06-13
Convert step-by-step instructions to use perl instead of sed

robinwhittleton:
GNU sed isn’t posix compatible, so fails on bsd systems. We can use perl instead (enforcing utf8 where necessary). Also add a warning to check roman numerals.

I’ve left the sed usage in the deploy-ebook-to-www script as that’ll only realistically be run from GNU machines.

Fixes https://github.com/standardebooks/web/issues/136.


acabal:
Sorry, I had to force push after this was submitted and screwed up this pull request. Can you rebase it against master and try again?

robinwhittleton:
No problem, I think that does it. I’m not up on the differences between single- and double- quoting in shell scripts, but I left your sed comment in place for the perl version.

acabal:
Thanks!

PR #144, opened 2021-07-22
Add review guide

thewchan:
First time trying to make a direct change to the website. Please let me know if anything is out of order.

### DIFF
--- a/www/contribute/how-tos/index.php
+++ b/www/contribute/how-tos/index.php
@@ -14,6 +14,9 @@ require_once('Core.php');
 		<li>
 			<p><a href="/contribute/how-tos/illustrations-music">How to create figures for music scores</a></p>
 		</li>
+		<li>
+			<p><a href="/contribute/how-tos/review">How to review a production before it is published</a></p>
+		</li>
 	</ul>
 </article>
 </main>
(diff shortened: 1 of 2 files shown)


acabal:
Great work Matt, thanks!

Some suggestions:
- The headers of `Step N` are redundant because the steps are already numbered with `<li>`. Can you think of some more descriptive headers?
- Step 1: Making sure the toolset is up to date should be outside of the steps, probably in the intro paragraph.
- For all the command line commands, can you add complete syntax highlighting a la the step by step guide? For example paths (including `.`) are underlined, strings are orange, globs are blue, etc.
- Step 9: Can you include an example of what a possible change might be? This is looking for errors like `pork ‘n’ beans`.
- All `<li>` items should have at least one block level child, usually `<p>`.

Thanks!

thewchan:
@acabal will do; I'll incorporate this sometime this week. Thanks!

thewchan:
@acabal Do you have a reference for the syntax highlight vs. the proper `class` attribute to use for the right css style/color? If not I can make one, I'm thinking that would be helpful to have in the readme of this repo

acabal:
No, IIRC it was auto-generated by Pygments or something like that. Just look at similar commands in the step by step guide and copy and paste whatever style is in there.

thewchan:
Got it; should I make a PR to the readme though once I figured out the style? Or do you think that's superfluous?

acabal:
No, since it was generated by a program it's better to just copy and paste for now.

thewchan:
@acabal 
I've addressed each points. Let me know if there are any other things. Thanks!

acabal:
Excellent, thanks!

Issue #140, opened 2021-06-27
Navigation options for web book text


robinwhittleton:
Search engines are indexing the text and dropping people onto chapters or other parts of books. For example, I just ended up on https://standardebooks.org/ebooks/evelyn-underhill/practical-mysticism/text/halftitlepage

Do we want/need some sort of mini navigation to rescue people who can’t use the back button to proceed?

acabal:
Maybe, but what the heck did you search for that it returned a top result of a page with only 8 words in it?

The nice thing about the current system is that we can just explode the epub and then expose it to the web. But if we added navigation then it would be a lot more post processing. Plus, if we add more elements then their styling could be messed up by the book CSS. For example what would happen to a top nav bar for a dedication, where children of `<body>` are centered using flex?

robinwhittleton:
That specific example came up in a search for pages that mention `"standardebooks"` that were new in the last week. So not a normal query 😁 But still, I can easily imagine a user searching for a quote and ending up on one of our pages. It’s maybe not worth the hassle though.

If browsers supported `<link rel="prev"/>` / `<link rel="next"/>` I’d suggest adding them to the head, but that never happened.

Close if you feel like it’s not useful.

acabal:
`<link rel>` would be a good addition I think. Another option is to use robots.txt to block indexing of the books themselves. Though I'm not sure if that's really what we want to do. If you want you can put together a PR to include `<link rel>` in the head in `deploy-ebook-to-www`. Relevant code starts around line 249.

acabal:
Though it may be tough to do it in Bash since you'd have to parse the spine. Maybe it's time for `deploy-ebook-to-www` to be rewritten in Python. Though I don't have time for that for the foreseeable future!

acabal:
I've thought about this a little more, and looked at some of the search results that lead people to SE, and it does look like Google et al are indexing at least the ToCs of a lot of the ebooks and sending people there for certain search queries. So maybe it does make sense to inject a header into the top of each exploded epub. Do you want to explore some possibilities for that?

Personally I think sticky headers are very annoying, but they do have some limited usefulness for small mobile devices. Maybe a small header with our name/logo that links to the homepage, and then a link to the ebook ToC? The header might be non-sticky for large screens but sticky for ipad size and below?

[...]

PR #146, opened 2021-07-26
Add a simple nav bar to web versions of books

robinwhittleton:
This is a really simple nav bar that just sits at the top of the page on desktop, but on portrait iPad and lower will remain fixed in place. The nav bar has two options: an SE logo that goes back to the homepage, and a “Table of contents” link that goes to the ToC. For single-page, we drop the ToC link as the user can just scroll to the top of the page anyway.

**Desktop view, showing fixed nav:**

Screenshot 2021-07-26 at 21 22 42

**Mobile view, showing docked nav:**

Screenshot 2021-07-26 at 21 24 51

**Single page view, with no link to ToC:**

Screenshot 2021-07-26 at 21 25 20

**Dark mode view:**

Screenshot 2021-07-26 at 21 25 48

This PR also updates the Dockerfile to get `deploy-ebook-to-www` working, and fixes a bug with missing nofollows on single-page sponsored colophons.

Fixes https://github.com/standardebooks/web/issues/140.

### DIFF
--- a/scripts/deploy-ebook-to-www
+++ b/scripts/deploy-ebook-to-www
@@ -320,7 +320,8 @@ do
 		find "${workDir}"/src/epub \( -type d -name .git -prune \) -o -type f -name "*.xhtml" -print0 | xargs -0 sed --in-place --regexp-extended "s|<title>|<title>${workTitle} - |g"
 
 		# Add the nav to each page
-		find "${workDir}"/src/epub \( -type d -name .git -prune \) -o -type f -name "*.xhtml" -print0 | xargs -0 sed --in-place --regexp-extended 's/<body(.*?)>/<body\1><header><nav><ul><li><a href="\/">Standard Ebooks<\/a><\/li><li><a href="\.">Table of contents<\/a><\/li><\/ul><\/nav><\/header>/'
+		bookUrl=$(grep --only-matching --extended-regexp "<dc:identifier id=\"uid\">.+?</dc:identifier>" "${workDir}"/src/epub/content.opf | sed --regexp-extended "s/.*?url:(.*?)<.*/\1/g")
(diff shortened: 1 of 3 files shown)


robinwhittleton:
Forgot to say: please can you test this locally before merging? It seems to work, but the `sed` stuff is out of my comfort zone.

acabal:
Looks very nice!

I screwed something up when I added a commit here. Can you force push from local and just make the tweak that I did in that commit, which fixes a broken selector.

Some suggestions:
- The link to the ToC is `.`. This works for most pages but if we're currently on the ToC, then it will actually to go up to the ebook detail page. We should use a relative full URL instead instead of an "up" URL.
- The size of the logo is a little small on my big monitor, hard to read. We can probably make it about 2x the size of the ToC link to better balance visually. You can also size it with rems so that it resizes when the page font size changes.
- You can "hide" the text in the logo link using `font-size: 0`. This is better than negative text indent because it actually disappears the text instead of just moving it really far off.
- The header uses position absolute, which takes it out of the document flow when calculating margins and such. So, we need to add extra margin-top to `body` to get the same amount of margin that we started with before the header bar was added.

robinwhittleton:
The styling tweaks have been done. Regarding the relative URLs, I can’t find a way to access the absolute URL from within the script. How about removing the ToC link on the ToC page instead?

acabal:
Since the script can look at content.opf, just get the dc:identifier with grep. dc:identifier is always the full URL of the details page.

robinwhittleton:
OK, I think this covers all your suggestions now. I couldn’t think of a particularly good wording for the link back to the details page, so I went with `Back to ebook`; happy to amend of course if you can think of better wording.

acabal:
Excellent, thanks! This will require a rebuild of the corpus so I'll do that later this week probably.

PR #148, opened 2021-10-01
Update step-by-step guide to mention PG copyright issue

weijia-cheng:
Given the recent discussion on the Google Group about *Twenty Thousand Leagues Under the Seas* and the copyright problem, I figured we should update this part of the guide to call that out as an example of what *not* to do.


acabal:
I agree, but I think the rewrite takes the focus away from the main point (PG may host different versions of the same book, so the producer must compare the all to select the best one and not necessarily the most popular one) and makes the focus the mistaken copyright status. I think the approach we should take instead is to select a different example than 20,000 leagues. What other books might have multiple translations available, where one is clearly superior?

weijia-cheng:
How about *The Art of War*? We use [Giles's version with annotations](https://www.gutenberg.org/ebooks/132), which is clearly better than [the incomplete version without annotations](https://www.gutenberg.org/ebooks/17405), though the best one here is also the most popular one so it's not quite the same.

weijia-cheng:
Hmm, actually a better example might be *A Journey to the Center of the Earth* since we use the [less downloaded, more original version](https://www.gutenberg.org/ebooks/3748) over the [more popular rewritten version](https://www.gutenberg.org/ebooks/18857).

weijia-cheng:
I've updated the text to give both a good example of picking a PG version (_Journey to the Center of the Earth_) and a bad example (_Twenty Thousand Leagues Under the Seas_).

acabal:
Thanks!

PR #149, opened 2021-10-02
Wanted: When We Were Very Young, by A. A. Milne

bentley:
It’s typical for the original *Winnie the Pooh* to be sold in a box set with *When We Were Very Young*, *Now We Are Six*, and *The House on Pooh Corner*.

P.S.: I mentioned this book on the mailing list, but my message never showed up. Since I self‐host my mail server, I’m always conscious about being filtered as spam. Could you check the Google Groups moderation queue?


acabal:
Good idea, thanks!

PR #150, opened 2021-10-16
Suggested process improvement for step-by-step guide: Separate initial commit from rough cleanup

weijia-cheng:
I believe that the existing step to create the initial commit should be separated into two commits, one to add the source text as-is and another to remove the cruft from `body.xhtml`. My reasoning is that it is quite possible for an overzealous trimmer to overshoot and delete real content from the book while deleting cruft. Then, a reviewer would not be able to catch this, since the commit history will start with those deletions already in place. Having a separate "delete cruft" step will make it easier for reviewers to know whether cruft was deleted correctly at the start of the project.


weijia-cheng:
That makes sense. Glad to learn the context.

PR #152, opened 2021-12-28
Flesh out step-by-step guide

weijia-cheng:
Added the full copy-and-paste-able commands for the three `find-*` steps.


acabal:
Thanks!

PR #156, opened 2022-01-06
Explain how to include translator(s) in step-by-step guide

weijia-cheng:
This got me on my first first production (_Penguin Island_) so it feels good to circle back to this.


acabal:
Thanks!

Issue #154, opened 2022-01-04
Front page should not advertise illustrations


bentley:
[devices.png](https://github.com/standardebooks/web/blob/master/www/images/devices.png) and [devices@2x.png](https://github.com/standardebooks/web/blob/master/www/images/devices%402x.png) depict illustrations within our edition of *Treasure Island*. I think this causes some confusion for people less familiar with the current SE policy on illustrations, and gives them certain expectations that result in wordy discussions on the mailing list.

Speaking for myself, it did give me the wrong impression of the illustrations policy when I first got involved.

acabal:
Yes, I suppose we should change that. I don't have time to at the moment, so if anyone is interested please jump in.

PR #157, opened 2022-01-06
Update homepage devices image to remove colour illustrations

robinwhittleton:
This uses The Return of Sherlock Holmes instead, which allows us to demo that we still have black and white illustrations where necessary. I also updated the generic android tablet on the left to demo a full cover, and added AVIF versions.

The new devices image.


robinwhittleton:
Fixes https://github.com/standardebooks/web/issues/154.

acabal:
Excellent, thanks!

PR #158, opened 2022-01-07
Change devices picture to use width instead of density

weijia-cheng:
Something I just realized after the devices image change is that the devices has always rendered very poorly on my laptop, appearing in low resolution. This is what the front page looks like on my laptop without this PR (you will need to see the full-size image to see the difference):

Screenshot 2022-01-06 at 21-27-16 Standard Ebooks

Doing some research, I found that this is because the `<srcset>` attribute for the the devices image uses the pixel density to decide what image to display instead of the image width.

```
<source srcset="/images/devices@2x.avif 2x, /images/devices.avif 1x" type="image/avif"/>
<source srcset="/images/devices@2x.png 2x, /images/devices.png 1x" type="image/png"/>
```

To put this in more concrete terms, my laptop has a pixel ratio of 1 (according to mydevice.io) so it pulls up the 670 pixel-wide 1x version and upscales it to a width of `40rem` (in my case 880 pixels, hence the blurriness).

Pixel ratio isn't a good criteria for this image, because unlike the other images on the page the devices image is scaled up or down depending on the width of the screen. ([MDN](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images) suggests that x-descriptors are appropriate when everyone sees the image at the same real-world size on the screen, which isn't the case here. I see the image at a considerably larger size on my laptop than on my phone.)

The solution I found was to change the `<srcset>` attribute to use width instead of pixel density, which has the following effect:

Screenshot 2022-01-06 at 21-43-05 Standard Ebooks


acabal:
Good catch, thanks!

Though, wouldn't a more complete solution be to simply correctly size the image to begin with?

That (original) image was one of the first things I put together for the SE site. What probably happened was that it was at some original resolution that was correct, but later on the max width of the container changed, and the image didn't. So it got stretched.

@robinwhittleton can we resize the new image to be 40rem in width? I.e. 880px@1x?

While we're on the subject, I just noticed that the "Dancing Men" header is not next to the "Dancing Men" illustration. Would it be too much work to move the "Dancing Men" header to the 2nd-from-last ereader, so that it's next to its illustration?

weijia-cheng:
Ok, resizing it makes sense. I'll let you or @robinwhittleton make the change since you probably have a better idea of how to handle this.

robinwhittleton:
No problem, let’s just remake the image at 880px, single density. Alex: thanks for the graphics, but probably best to update to modern devices anyway. So I’ll go high-end Kobo, Kindle Paperwhite, iPad and generic Android tablet again.

acabal:
We do need it at 2x too, but since we're at it, can this be an SVG?

robinwhittleton:
New devices. Look OK?

devices@2x

I can export this as an SVG but it’s coming up at 3.3MB. I’m not sure Pixelmator’s SVG export is particularly clever. We could remake it do just be vector frames and actual text, but that’d take more time than I’m willing to throw at it unfortunately.

acabal:
Looks good, but can you move the dancing men header next to the reader when the actual dancing men?

I was thinking starting with SVG frames and converting the reader contents to SVG, but if you don't have time that's fine, this looks good too

On January 9, 2022 10:29:18 AM CST, Robin Whittleton ***@***.***> wrote:
>New devices. Look OK?
>
>
>
***@***.***(https://user-images.githubusercontent.com/7414/148691220-8c0b0469-7b9f-4baf-9d94-3c1b65f4174e.png)

acabal:
One last thing-- once you finish, can you email me the source files privately so I can keep them in a repo in case we need to make changes later?

robinwhittleton:
PR for that at https://github.com/standardebooks/web/pull/159

PR #159, opened 2022-01-09
Update devices image to higher res and newer devices

robinwhittleton:
Devices are now iPad Pro, Kindle Paperwhite and Kobo Sage. The image has been rebuilt from scratch to a higher default resolution of 880px (1x).


acabal:
Great work! There are a few white pixels at the upper right that look off when viewed in dark mode, can you take care of those?

robinwhittleton:
Ffs, sorry!

PR #162, opened 2022-01-26
Allow Safari to rescale the SE logo.

tuskpot:
This is related to the logo scaling issue in Safari that was reported on the mailing list here:
https://groups.google.com/g/standardebooks/c/PusDL5_DSEc

This is one possible fix. This simply adds `preserveAspectRatio="none"` to the svg. Without this, Safari refuses to change the aspect ratio of the svg.

I don't know if this change will have side effects elsewhere in the site or the corpus.

There's an open bug report (since 2012) for this issue on the webkit bug tracker:
https://bugs.webkit.org/show_bug.cgi?id=82645


acabal:
OK, this works. I don't know if it would have a side effect either, so what I'm going to do is just copy the SVG into a different file just for the spine. We can make it white which will also allow us to remove the invert filter on the spine, simplifying things a lot. Thanks for figuring this out! Done in 450aa5b

PR #163, opened 2022-01-26
Cleaning up wanted books

mgbennet:
Remove several books already in the collection, and one duplicate entry.


acabal:
Good catches, thanks!

PR #165, opened 2022-03-15
Link to latest version of manual

ConcaveTrillion:
Instead of pointing to older version of manual, link to latest version


acabal:
We don't link to `latest` on purpose because the numbering can change between versions. The only sane way to make sure documents like these stay *correct* (if not *current*) without our having to review them constnatly is to link to specific manual versions.

ConcaveTrillion:
Makes sense. Can I submit a PR to update these specific links to the new version of the manual, or do you do a review every so often to get them all at once?

PR #167, opened 2022-04-17
Increase JPEG quality of 242px thumbnails to 85%

bentley:
[I’ve whined before](https://github.com/standardebooks/tools/issues/405) about how *The Great Gatsby*’s cover is badly affected by JPEG artifacts in the “Browse Ebooks” page. This one really sticks out like a sore thumb…nail. It may affect others with similar characteristics too (solid color with scattered details?).

I’d like to propose bumping the quality on these smallest images to 85%, the maximum quality suggested by Google Page Speed Insights. It makes the artifacting a lot less noticeable, and in my simplistic tests increases the filesize by about 25%, or under 2 kilobytes per thumbnail. That seems negligible to me—although of course I’m not the one who has to cover bandwidth and placate PageRank.

90% provides visibly better results but more than doubles the filesize, so that seems not so useful.

### DIFF
--- a/scripts/deploy-ebook-to-www
+++ b/scripts/deploy-ebook-to-www
@@ -226,7 +226,7 @@ do
 		sed -i "s/cover\.jpg/${urlSafeIdentifier}\.jpg/g" "${imgWorkDir}/${urlSafeIdentifier}.svg"
 
 		# Resize and compress the image (formula from Google Page Speed Insights)
-		convert -resize "242" -sampling-factor 4:2:0 -strip -quality 80 -colorspace RGB -interlace JPEG "${imgWorkDir}/${urlSafeIdentifier}.svg" "${imgWorkDir}/${urlSafeIdentifier}-cover.jpg"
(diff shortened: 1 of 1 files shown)


acabal:
Does a 5% increase actually make a visible difference? The Gatsby thumbnail is the most noticeable because its composition is the kind of image that JPG is traditionally bad at: small-high contrast detail of the lips against the sky. At the moment I don't think any other thumbnail looks nearly as bad because most other ones don't have that kind of composition. And I imagine that 5% isn't going to look a whole lot better given this poorly-suited image.

bentley:
Yes, I think the difference is visible, especially in the avif.

80%: 
thumb-80 avif

85%: 
thumb-85 avif

90%: 
thumb-90 avif

At 80% the top edge of the artifact manifests a light‐colored thin line, and the bottom edge is straight, making the lips look like a single box‐shape. At 85% there is still some artifacting but the top and bottom edges are close to the background color, so it’s no longer obviously square.

acabal:
I see a difference but honesty I don't think it's a very big one, and 85% still has clear artifacting around the lips. I don't think it's worth a 25% increase in file size especially since it's really only Gatsby that has this problem. We'd be increasing total bandwidth cost by 25% across the whole corpus, to only slightly improve a single image, and not even actually fix it because it would still be artifacted. So let's put this aside for now, sorry :(

PR #168, opened 2022-04-23
Add note to run 'se clean' when completing content.opf

vr8hub:
I know the manual mentions that `se clean` will escape the HTML in long description, but since it came up today in a review, I thought maybe it was worth mentioning it needed to be run in the content.opf section of the Step-by-Step guide?

If not, no biggie.


acabal:
OK, but we should word it to work in cases where the author composed escaped html to begin with. Something like, 'If you wrote the long description in unescaped html...'

vr8hub:
lol. I thought about that, but since you left if lowercase I thought that's the way you preferred it. I didn't even notice that I capitalized it at the end of the sentence. Doh!

Done.

acabal:
Thanks!

PR #169, opened 2022-04-29
Make italics review step command use less as a pager

weijia-cheng:
Unlike `git grep`, `grep` doesn't page by default, so I added a pipe to `less` to get it to page.


acabal:
I don't want to do this because there are other git pagers out there that may conflict with having `less` in copy and paste. For example in my git config I have `pager = diff-so-fancy | less --tabs=4 -RF` and changing the copy and paste results in no colors for me.

PR #173, opened 2022-05-22
Add punctuation to example interactive-replace

irontigran:
Add exclamation mark and question mark to the interactive-replace command as punctuation that should be inside quotation marks.

Ran across question marks a number of times working on Plato, this seemed like a worthwhile update to the step-by-step page.


acabal:
Only commas and periods go inside quotation marks - exclamation marks and question marks generally don't (unless it's part of a quotation).

irontigran:
Hmm, my mistake on that then. I over-generalized the question marks that needed to go in the quotation.

Issue #174, opened 2022-05-28
Are find-* changes editorial?


vr8hub:
(I was going to do a PR highlighting editorial commits before I saw you'd just done one. I think the last four reviews I've done have all had to be rebased. So thanks for that! I don't know if might be worth adding something to the boilerplate you include in the email to first-time producers, too?)

Are changes made for find-mismatched-dashes and find-mismatched-diacritics considered editorial? I assume so since they're changing content, but I don't think I've ever heard officially. If they are, I'll do a PR to add a note to that effect to each of them in the Step-by-Step.

While I'm here, having nothing to do with the above, what do you think about changing the Step-by-Step to add a commit right before typogrify that includes the splitting of the files and the clean, so that the typogrify commit becomes the third one? Because the guide includes a manual typography checklist, it's possible for people to make manual changes before that second commit. Since that commit also includes splitting the files _and_ clean _and_ what typogrify does, it makes it very difficult to pick out manual changes.

That's the minimum suggested change. In the "commits are free" vein, and in the vein of trying to keep commits focused on a single type of change, what I'd really suggest is to separate automated commits from manual commits in the Step-by-Step, and to include commits in a couple of places we don't currently have them. I believe this would make reviews of first-time productions a lot easier (reviews in general, but especially first-time ones), and it's only a few extra commits, so I don't think it would over-burden a new producer. 

(I'm just labeling what it's in them, not necessarily giving the commit message we might want)
Initial commit (step 5)
_Split/clean commit_ (step 7)
Typogrify commit (step 8a)
_Manual typography changes commit_ (step 8b)
_Footnote-to-endnote commit_ (step 10) (not needed that often, but if it is, IMO it should be its own commit)
British to American commit (step 11)
Semanticate commit (step 12a)
_Manual semantic changes commit_ (step 12b)
Modernize spelling commit (step 13)
_Find-mismatched-* commit_ (step 14/15) (could be one, or one each)
Manifest/spine commit (step 17)
ToC commit (step 18)

That's five new commits (in italics), but one of them isn't used that often. I think it would also help first-time producers get in the habit of isolated commits, i.e. not too many changes in a single commit.

Just throwing it out there. I think it would have helped me with my first couple of books, where I wasn't used to git and had far too few commits. And, as I said, I think it would definitely help reviewers.

I'll do a PR for whatever you think is worth adding, if anything.

acabal:
[...]

PR #175, opened 2022-05-30
Add additional commits

vr8hub:
Per discussion on #174. I'm afraid I've never gotten the website working; it hasn't been a big deal to date because my web changes have always been small. This is obviously larger, and it's certainly possible (probable?) I messed something up. I would normally try again to get it up and running before submitting this PR, but we haven't seen our kids in two years and they just got in today, so writing a little text after they went to bed wasn't a big deal, but trying to get a VM _and_ the web site up and running is out of scope right now. :)

All of that to say, if there's a problem, just leave it and I'll try again to get a VM working in a couple of weeks after they leave (for a bit). I thought it was worth going ahead and posting on the off chance I _didn't_ mess anything up. :) But I don't expect you to fix it if I did, so no worries if it has to wait a while.


acabal:
Great work, thanks!

PR #177, opened 2022-06-17
Add guide to common issues in converting from PD transcriptions

vr8hub:
All right, here is the first pass. I did this originally as an .rst file based on the manual, which made the classes all wrong for something like the step-by-step. So I mostly did this by hand, using the Step-by-Step as a template of sorts. I'm sure there are mistakes; let me know what needs to be fixed and/or reworded.

### DIFF
--- a/www/contribute/common-issues-converting-from-pd-transcriptions.php
+++ b/www/contribute/common-issues-converting-from-pd-transcriptions.php
@@ -8,30 +8,30 @@ require_once('Core.php');
             <li>
                 <p>Punctuation, other than periods, appearing immediately inside a closing parenthesis should be moved outside the parenthesis.</p>
                 <p>This comma that is inside the closing parenthesis…</p>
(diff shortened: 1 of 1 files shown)


acabal:
Thanks Vince. There's a lot of broken XHTML in here, which is causing failed rendering. If you can't get the website to work locally, then I suggest saving the non-PHP bits of this file into an XHTML file, then adding the headers and footers from the actual SE website which you can get by using view source. Then, open that XHTML file in your browser and it'll tell you where all the parse errors are.

vr8hub:
OK, will do.

vr8hub:
It was mostly just missing closing code tags. However, the last item, which is a blockquote, is not displaying correctly, even with all the CSS links in the header (copied from the Step-by-Step page). I have it formatted like the formatted HTML in the Step-by-Step, e.g. using tags at the beginning of the lines for indenting. If I need to do something different, let me know.

acabal:
Great, thanks!

PR #180, opened 2022-07-02
How to review guide: Remove duplicated instruction

weijia-cheng:
This instruction is already given in more detail on line 62.


weijia-cheng:
I also found another issue, the command `grep --recursive --line-number "<i|<em" src/epub/text/*.xhtml` should have the `-E` parameter since the `|` needs extended regex support.

PR #183, opened 2022-07-25
Fix typo in How to Use Our Ebooks

bearmarch:
imgae -> image in How to Use Our Ebooks -> Method 2 - Download Directly to Kindle


acabal:
Thanks!

PR #185, opened 2022-07-27
Restrict Internet Archive source to scans only

weijia-cheng:
We have been using the Wayback Machine sometimes to create a stable link to a transcription if it has gone down in the past, e.g. [*Lyrical Ballads*](https://standardebooks.org/ebooks/william-wordsworth_samuel-taylor-coleridge/lyrical-ballads). The logic to categorize sources on the website lumps all Internet Archive sources as scans. It should only treat an IA source as a scan if it has the `/details` path (this is also consistent with our lint rule for Internet Archive paths).


acabal:
Thanks!

PR #187, opened 2022-07-31
Remove In Search of the Castaways from wanted list

SzymonSzott:
It has already been produced: https://standardebooks.org/ebooks/jules-verne/in-search-of-the-castaways/j-b-lippincott-co


acabal:
Thanks!

PR #188, opened 2022-08-10
Tweak Epictetus suggestion

weijia-cheng:
I suggest that the final resolution to the Epictetus problem is just to use Higginson's translation. As he says in his [preface](https://babel.hathitrust.org/cgi/pt?id=hvd.32044085112423&view=1up&seq=11&skin=2021) it's basically a revision of Carter's translation with more modern language. It has both the Enchiridion and the Fragments so completeness shouldn't be a problem. It also has a transcription so it doesn't need to be an advanced project.


acabal:
Thanks!

Issue #189, opened 2022-08-22
List of Authors


donatj:
https://standardebooks.org/feeds/atom/authors

So this page has the full list which I only found after poking around the codebase for a minute.

It'd be super helpful to have like an "Authors" link in the header or footer - or like a dropdown of authors on the Books page?

I'd even happily donate the time to build it if I got a little direction ;)

acabal:
There's a link in the site footer to https://standardebooks.org/feeds which lists all the available feeds. You can drill down from there to see the different feed options for your preferred feed format. Does that clear things up?

donatj:
No. I'm not looking for feeds.

I'm looking for a way to find books by _specific authors_ from the website. Right now if you're looking for books by an author, you have to paginate through _all the books_ until you find one by them and then click on their name - which takes you to https://standardebooks.org/ebooks/apuleius

I want a way, from the website to find an author without paginating through books.

acabal:
In that case we'd need some help doing that. Putting an index in `/authors` would probably be the way to go about that. `/authors/<AUTHOR_SLUG>` could be an Apache internal redirect to `/ebooks/<AUTHOR_SLUG>`, preserving the URL but resusing the author listing code.

acabal:
If you'd like to work on this, then please go ahead. Let me know if you have any specific questions as you get started!

donatj:
Yup, I will poke at it ;)

PR #191, opened 2022-08-24
Add authors page

donatj:
Finishes #189 

Here's the basic authors page I was talking about.

image

I don't have the full set populated so my screenshot is limited.

I'm more than happy to change anything. I tried my best to match the pre-existing coding style.

### DIFF
--- a/lib/Library.php
+++ b/lib/Library.php
@@ -487,12 +487,15 @@ class Library{
 				}
 
 				// Create the authors cache
-				$authorPath = EBOOKS_DIST_PATH . rtrim(preg_replace('|^/ebooks/|ius', '', $ebook->AuthorsUrl), '/');
-				if(!array_key_exists($authorPath, $authors)){
-					$authors[$authorPath] = [];
+				$authorPaths = [ EBOOKS_DIST_PATH . rtrim(preg_replace('|^/ebooks/|ius', '', $ebook->AuthorsUrl), '/') ];
+				foreach($ebook->Authors as $ebookAuthor) {
+					$authorPaths[] = EBOOKS_DIST_PATH . rtrim(preg_replace('|^/ebooks/|ius', '', $ebookAuthor->UrlName), '/');
 				}
 
-				$authors[$authorPath][] = $ebook;
+				$authorPaths = array_unique($authorPaths);
+				foreach($authorPaths as $authorPath) {
+					$authors[$authorPath][] = $ebook;
+				}
 			}
 			catch(\Exception $ex){
 				// An error in a book isn't fatal; just carry on.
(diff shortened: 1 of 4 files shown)


acabal:
This is great, thanks! Some notes:

- We should sort authors by their sort name, not their regular name. Also note that `<=>` is not Unicode aware so it will sort names like `Karel Čapek` incorrectly. You should sort using a collator, see https://github.com/standardebooks/web/blob/f9665ee339a93d0774012df51aac757e536b60f3/lib/Library.php#L368 for a pattern to follow.
- When running on my machine, PHP gives an error that `$authors->AuthorsUrl` is undefined. Instead you should do `<?= '/ebooks/' . $author->UrlName ?>`.
- Some ebooks can have more than one author, like https://standardebooks.org/ebooks/karl-marx_friedrich-engels/the-communist-manifesto/samuel-moore so you should use `strip_tags($ebook->AuthorsHtml)` to get the author "name" instead. However you can *sort* that name by using `$ebook->Authors[0]->SortName`.

Some testing books for you to add are:

-https://standardebooks.org/ebooks/karel-capek/r-u-r/paul-selver_nigel-playfair
-https://standardebooks.org/ebooks/karl-marx_friedrich-engels/the-communist-manifesto/samuel-moore
-https://standardebooks.org/ebooks/samuel-butler-1612-1680/hudibras (note the strange author slug, to differentiate between another author named `Samuel Butler`.

donatj:
@acabal - [Here's my proposed solution](https://github.com/standardebooks/web/pull/191/commits/1157d02fa0842ec2febc70b6b5d820a4b6882e21) - I'm happy to revert it if it doesn't work for you.

Essentially we create pages for the _individual authors_ in addition to the group pages. 

This way, if we have someone who wrote a book individually and as a group their individual author page would list both.

donatj:
Just incase it wasn't obvious, you have to rebuild your book cache to test my code here.

acabal:
I don't think we're ready yet to split multi-author works like Communist Manifesto into their individual authors quite yet. This is because bulk downloads/feeds are grouped by author group and not individual authors, and the author links themselves in an ebook page (i.e. clicking on the author name below the book title in the ebook detail view) leads to the group author page too. Group authors are treated like one unit across the entire site. 

If think it's a good proposal to split them up, but if we do that then we should do it across the site as a whole for consistency. I think that's out of scope for this PR right now.

Also, you've added `$collator` but you didn't define or instantiate it. How can this run on your machine without crashing?

acabal:
Closed due to inactivity. If you have time to wrap this up feel free to reopen the PR. Thanks!

PR #192, opened 2022-08-24
Update wanted ebooks list: Two Years Before the Mast

bentley:
Wikipedia: [Book](https://en.wikipedia.org/wiki/Two_Years_Before_the_Mast), [Author](https://en.wikipedia.org/wiki/Richard_Henry_Dana_Jr.)

In terms of notability, it was included in Harvard Classics, was adapted into a popular film, and Herman Melville (a contemporary) spoke highly of it.

In terms of complexity, there are some poems and footnotes. PG has a biographical preface that can be dropped, and a labeled diagram of a ship that could be kept or could be dropped. Typography will need some care: replacing ASCII apostrophes with primes in coordinates; replacing asterisked elisions with three‐em dashes; missing italics (plenty of ship names). 167,000 words.


acabal:
Thanks!

PR #194, opened 2022-09-18
Fix AVIF images in iOS16 (and 8+ times speedup of AVIF generation)

robinwhittleton:
iOS16 upholds AVIF standards more strictly than other renderers, which means that the current AVIF images on SE are broken. A fix for generation bug has been added to go-avif (https://github.com/Kagami/go-avif/pull/17) but we can do better.

This PR switches out the AVIF generation to cavif (https://github.com/kornelski/cavif-rs) which not only doesn’t suffer from the iOS16 but drops generation time by more than 8 times. For example, here’s a set of three timed runs in both go-avif and cavif on the title card of https://standardebooks.org/ebooks/jonathan-dymond/an-inquiry-into-the-accordancy-of-war-with-the-principles-of-christianity:

```
root@f58fb0beac90:/standardebooks.org/web/scripts# time ./go-avif -e "illustration.jpg" -o "illustration-go-1.avif" --best

real	0m31.191s
user	1m39.113s
sys	0m1.662s
root@f58fb0beac90:/standardebooks.org/web/scripts# time ./go-avif -e "illustration.jpg" -o "illustration-go-2.avif" --best

real	0m33.097s
user	1m49.846s
sys	0m0.831s
root@f58fb0beac90:/standardebooks.org/web/scripts# time ./go-avif -e "illustration.jpg" -o "illustration-go-3.avif" --best

real	0m36.963s
user	2m4.235s
sys	0m0.711s

root@f58fb0beac90:/standardebooks.org/web/scripts# time ./cavif --quality 50 "illustration.jpg" -o "illustration-cavif-1.avif"
illustration-cavif-1.avif: 98KB (96840B color, 0B alpha, 242B HEIF)

real	0m4.017s
user	0m11.600s
sys	0m0.051s
root@f58fb0beac90:/standardebooks.org/web/scripts# time ./cavif --quality 50 "illustration.jpg" -o "illustration-cavif-2.avif"
illustration-cavif-2.avif: 98KB (96840B color, 0B alpha, 242B HEIF)

real	0m4.083s
user	0m11.857s
sys	0m0.091s
root@f58fb0beac90:/standardebooks.org/web/scripts# time ./cavif --quality 50 "illustration.jpg" -o "illustration-cavif-3.avif"
illustration-cavif-3.avif: 98KB (96840B color, 0B alpha, 242B HEIF)

real	0m4.237s
user	0m12.339s
sys	0m0.093s
```

The executable has been generated via cargo, and has the following qualities:

```
root@f58fb0beac90:/standardebooks.org/web/scripts# file cavif
cavif: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=f972071996c58b7a855d04e4c5804f4a3d94f575, for GNU/Linux 3.2.0, with debug_info, not stripped
```

I’m not sure how cavif’s quality 50 matches go-avif’s best flag, but they produce similar file sizes and similar visual quality.

There is a slight different in apparent colour space. If this is an issue there are tweaks, but according to the documentation it makes files larger so should be avoided. Given that the titlecards will matches the covers I see no reason to go down that path.

**N.b. I haven’t tested this in situ with the corpus. @acabal could you find time to check the branch out and see what you think?**

### DIFF
--- a/scripts/deploy-ebook-to-www
+++ b/scripts/deploy-ebook-to-www
@@ -131,7 +131,7 @@ require "git" "Try: apt-get install git"
 require "php" "Try: apt-get install php-cli"
 require "se" "Read: https://standardebooks.org/tools"
 
-# go-avif is compiled from https://github.com/Kagami/go-avif/
+# cavif is compiled via Cargo: https://github.com/kornelski/cavif-rs
 
 scriptsDir="$(dirname "$(readlink -f "$0")")"
 
@@ -233,8 +233,8 @@ do
 		convert -resize "1318" -crop "1318x439+0+659" -sampling-factor 4:2:0 -strip -quality 80 -colorspace RGB -interlace JPEG "${imgWorkDir}/${urlSafeIdentifier}.jpg" "${imgWorkDir}/${urlSafeIdentifier}-hero.jpg"
 		convert -resize "2636" -crop "2636x860+0+1318" -sampling-factor 4:2:0 -strip -quality 80 -colorspace RGB -interlace JPEG "${imgWorkDir}/${urlSafeIdentifier}.jpg" "${imgWorkDir}/${urlSafeIdentifier}-hero@2x.jpg"
 
-		"${scriptsDir}"/go-avif -e "${imgWorkDir}/${urlSafeIdentifier}-hero.jpg" -o "${imgWorkDir}/${urlSafeIdentifier}-hero.avif" --best
-		"${scriptsDir}"/go-avif -e "${imgWorkDir}/${urlSafeIdentifier}-hero@2x.jpg" -o "${imgWorkDir}/${urlSafeIdentifier}-hero@2x.avif" --best
+		"${scriptsDir}"/cavif --quality 50 "${imgWorkDir}/${urlSafeIdentifier}-hero.jpg" -o "${imgWorkDir}/${urlSafeIdentifier}-hero.avif"
+		"${scriptsDir}"/cavif --quality 50 "${imgWorkDir}/${urlSafeIdentifier}-hero@2x.jpg" -o "${imgWorkDir}/${urlSafeIdentifier}-hero@2x.avif"
 
 		# Build the cover image thumbnail
 		# We use JPG instead of SVG for several reasons:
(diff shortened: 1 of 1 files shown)


acabal:
Great! I'm rebuilding the corpus now with this update. Can you check A A Milne and see if it works?

acabal:
After testing this on the server, ca-avif is compiled to require glibc >= 2.29 but Ubuntu 18.04 only has 2.27. It doesn't look like there's a way to get 2.29 on 18.04 without compiling from source and risking breaking the system. So we need to think of a different avif encoder, or wait until we upgrade the server to the next LTS which will not be soon.

acabal:
Actually never mind, I was able to download a binary from the source git repo and it works fine. Can you check A A Milne again when you have a moment? Make sure to force a cache refresh in your browser.

robinwhittleton:
Good news: It’s works!

Bad news: it’s downloading both the JPG and AVIF for no apparent reason. I guess this is a Webkit bug.

Screenshot 2022-09-19 at 07 37 34

Either way, the images are loading and the AVIF generation time is better, so if you’re happy let’s merge this and I’ll look into why it’s pulling both.

acabal:
OK, I guess that bug should be reported to Webkit somewhere. I've merged this in, thanks!

robinwhittleton:
Narrowed the bug down to `picture` elements in a page that’s served as XHTML, so no idea how long that’s been a problem for us. Filed at https://bugs.webkit.org/show_bug.cgi?id=245411.

acabal:
Hmm. That could be a serious bandwidth problem. Is it only Safari or is it also Chrome?

It might be worthwhile to check in PHP if Safari is accessing the page and then serving as HTML if it is, until this gets fixed.

robinwhittleton:
Good question. I’ve just checked and it doesn’t seem to be a problem in Blink, just WebKit.

acabal:
Is it only ios or also regular safari?

robinwhittleton:
Both, and not just the latest versions. It’s a WebKit problem dating back presumably to the introduction of the picture element (beginning of 2016). I guess not many SE devs use Safari dev tools, and no-one noticed. And next to no-one serves their page as XHTML so it’s not been picked up in the wider community.

acabal:
OK, I've pushed a change that will not serve AVIF sources to Safari. Can you check the live site to see if's working and only serving JPG sources to Safari?

robinwhittleton:
So whatever you’ve changed has fixed it on the ebooks listing page, which slightly surprised me. But the covers are still downloading twice on the ebook page.

I say surprised, because the reduced testcases in the WebKit bug I filed showed that the bug wasn’t related to AVIF at all, but seemed to be a bug in serving picture elements with `application/xhtml+xml`. So I’m not sure how you fixed that 😅

acabal:
All I did was check for Safari, and if it is Safari then it doesn't write out the `<source>` element for AVIF.

Can you use the inspector to check which specific elements are still being downloaded twice? There should be no `<source>` elements with AVIF in Safari.

robinwhittleton:
There aren’t any AVIF images, but it’s not related to AVIF: the testcases in the bug I filed are just JPGs. It’s some bug with `picture` and XHTML. However, I don’t understand why the hero page is downloading both the JPGs and the list page is now just downloading one per item.

Are you worried about the bandwidth costs? Safari users tend to upgrade fairly quickly, so it might be more effective to target Safari 16 and up with just the AVIF image as the `` source with no `<source>` elements?

acabal:
Yes, bandwidth is the concern, we don't want Safari to be downloading the same image from multiple sources for no reason.

What's the solution here for Safari? Remove all `<source>` and just have `<picture></picture>`?

PR #202, opened 2022-11-05
Add chapter navigation footer

brandondong:
I use the web interface for reading and one personal gripe is the lack of previous/next controls when reaching the end of a chapter. Using the back button to go back to the table of contents works but I frequently access chapters from bookmarks.

I tried my hand at implementing such a feature and I believe I have everything working.

### Screenshot:
image

### Missing one of previous/next:
image

### Titles too long relative to the screen width:
image

### Dark mode:
image

### DIFF
--- a/scripts/deploy-ebook-to-www
+++ b/scripts/deploy-ebook-to-www
@@ -339,8 +339,11 @@ do
 		workTitle=$(grep --only-matching --extended-regexp "<dc:title id=\"title\">(.+?)</dc:title>" "${workDir}"/src/epub/content.opf | sed --regexp-extended "s/<[^>]+?>//g")
 		find "${workDir}"/src/epub \( -type d -name .git -prune \) -o -type f -name "*.xhtml" -print0 | xargs -0 sed --in-place --regexp-extended "s|<title>|<title>${workTitle} - |g"
 
-		# Add the nav to each page
+		# Add the header nav to each page
 		find "${workDir}"/src/epub \( -type d -name .git -prune \) -o -type f -name "*.xhtml" -print0 | xargs -0 sed --in-place --regexp-extended "s%<body(.*?)>%<body\1><header><nav><ul><li><a href=\"/\">Standard Ebooks</a></li><li><a href=\"${bookUrl}\">Back to ebook</a></li><li><a href=\"${bookUrl}/text\">Table of contents</a></li></ul></nav></header>%"
+
+		# Add a chapter navigation footer to each page
+		"${scriptsDir}"/inject-chapter-navigation-footer "${workDir}"/src/epub "${bookUrl}"
 
 		# Adjust sponsored links in the colophon
 		sed --in-place 's|<p><a href="http|<p><a rel="nofollow" href="http|g' "${workDir}"/src/epub/text/colophon.xhtml
--- a/www/css/web.css
+++ b/www/css/web.css
@@ -1,5 +1,9 @@
 @namespace epub "http://www.idpf.org/2007/ops";
 
+html{
+	position: relative;
+}
+
 body{
 	font-family: "Georgia", serif;
 	font-size: 18px;
@@ -49,6 +53,28 @@ body > header li:first-child > a{
 
 body > header li:first-child > a:hover{
 	transform: scale(1.025) rotate(1deg);
+}
+
+body > footer{
+	position: absolute;
+	bottom: 0;
+	left: 0;
+	width: 100%;
+}
+
+body > footer ul{
+	display: flex;
+	justify-content: space-between;
+	margin: 0;
+	padding: 0.5em 1em;
+	list-style: none;
+}
+
+body > footer li{
+	max-width: 40%;
+	overflow: hidden;
+	text-overflow: ellipsis;
+	white-space: nowrap;
 }
 
 body > section[epub|type~="titlepage"],
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ If everything went well you should now be able to open your web browser and visi
 mkdir /standardebooks.org/ebooks/
 cd /standardebooks.org/ebooks/
 git clone --bare https://github.com/standardebooks/david-lindsay_a-voyage-to-arcturus
-/standardebooks.org/web/scripts/deploy-ebook-to-www david-lindsay_a-voyage-to-arcturus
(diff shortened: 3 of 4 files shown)


acabal:
@robinwhittleton what do you think of this? Is this something we want?

robinwhittleton:
I think so!

Comments on the implementations:
- I think that rather than `<em>` we can just use `<i>`, it doesn’t need to be emphasised.
- I’m not a huge fan of `text-overflow: ellipsis` for accessibility reasons: as you bump the font size up you see less of the content. But as implemented we can’t avoid that. Potentially we should add a `<main>` around the rest of the content, apply the body margins to that directly, then we can just drop the `position: absolute`. That’ll allow the text to just flow onto multiple lines if needed.

brandondong:
### Screenshot

image

scripts/deploy-ebook-to-www
```diff
@@ -310,8 +310,11 @@ do
 			# We do this first because the tweaks below shouldn't apply to the single-page file
 			se recompose-epub --xhtml --output "${workDir}"/single-page.xhtml --extra-css-file="${webRoot}/css/web.css" "${workDir}"
 
+			# Wrap book contents in a <main> tag (and extra <div> for styling reasons)
+			sed --in-place --regexp-extended "s|<body([^>]*)>|<body><main\1><div>|; s|</body>|</div></main></body>|" "${workDir}"/single-page.xhtml
+
 			# Add a navbar with a link back to the homepage
-			sed --in-place --regexp-extended "s|<body(.*?)>|<body\1><header><nav><ul><li><a href=\"/\">Standard Ebooks</a></li><li><a href=\"${bookUrl}\">Back to ebook</a></li></ul></nav></header>|" "${workDir}"/single-page.xhtml
```
brandondong, line 314: sed doesn't actually support non-greedy regex matching (https://stackoverflow.com/questions/1103149/non-greedy-reluctant-regex-matching-in-sed) and so this would incorrectly insert the header after all the tags that follow body on the same line. Previously, body was the only opening tag on its line so this wouldn't matter.

www/css/web.css
```diff
@@ -3,9 +3,19 @@
 body{
 	font-family: "Georgia", serif;
 	font-size: 18px;
+	margin: 0;
+	min-height: 100vh;
+	display: flex;
+	flex-direction: column;
```
brandondong, line 9: Flexbox is necessary because the content of a chapter, like for the title page, may be quite short in height and we need the header to stick to the bottom (https://philipwalton.github.io/solved-by-flexbox/demos/sticky-footer/).

scripts/deploy-ebook-to-www
```diff
@@ -310,8 +310,11 @@ do
 			# We do this first because the tweaks below shouldn't apply to the single-page file
 			se recompose-epub --xhtml --output "${workDir}"/single-page.xhtml --extra-css-file="${webRoot}/css/web.css" "${workDir}"
 
+			# Wrap book contents in a <main> tag (and extra <div> for styling reasons)
```
brandondong, line 313: But one issue that using flexbox introduces is it removes the existing margin collapsing behavior of the body margins with the top + bottom section/image margins (https://stackoverflow.com/questions/43882869/margin-collapsing-in-flexbox). With the workaround, we get the old behavior back and everything should remain pixel identical with master excluding the new footer.

acabal:
Is there a way we can do this without a scaffolding `div`? If we're wrapping the book contents in `<main>`, why don't we do something like `justify-content` to center the text, we can remove `position: absolute` from the header, and then the top and bottom margins don't matter so much because flex aligns and spaces the three children of `<body>`. Does that make sense? I'm just going off the top of my head here, I haven't had time to experiment. But broadly speaking I think this should be possible without adding an extra div.

Also, make sure to test this on some books that use absolute positioning, because you may run in to strange results. [Cane is such a book](https://standardebooks.org/ebooks/jean-toomer/cane).

brandondong:
> If we're wrapping the book contents in `<main>`, why don't we do something like `justify-content` to center `<main>`, and then we can remove `position: absolute` from the header, and then the top and bottom margins of the text don't matter so much because flex aligns and spaces the three children of `<body>`. Does that make sense?

I don't think I completely follow the suggestion. Right now, main has `flex-grow: 1` so any value of `justify-content` will do nothing. If you remove `flex-grow: 1` and do something like `justify-content: space-between`, then the title page would show the image centered versus at the top where it is currently.

The reason I was worried about unintentionally removing margin collapse was because I thought that this footer change could not introduce any unrelated top/bottom whitespace changes (i.e. remain pixel identical with master). If this is not important, as it sounds like, then you could just reduce top margin from 8rem -> 5rem, probably leave bottom margin alone, and then call it a day as a best effort estimate to compensate for the removal of margin collapse.


> Also, make sure to test this on some books that use absolute positioning, because you may run in to strange results. [Cane is such a book](https://standardebooks.org/ebooks/jean-toomer/cane).

I've verified all pages render identically to master. There are some instances like https://standardebooks.org/ebooks/jean-toomer/cane/text/part-3 where the header will cut off the image. Likewise, if some image decides to do something like `bottom: 0`, then it would interact poorly with the new footer. With the new `main` element, there is an opportunity to fix this issue as well as the existing header issue by sticking a `position: relative` on `main` and making the header not `position: absolute`. But maybe this is a bad idea -- in the Cane example, it uses `top: 5vh` and `bottom: 5vh`, and `main` would be slightly less than the `100vh` that the author was envisioning when laying out elements.

acabal:
Maybe I meant `align-content` or `align-items` instead of `justify-content` when using `flex-direction: column`. Basically whatever the keyword is to achieve the equivalent of `margin: 0 auto` for centering a fixed-width column. 

We could also use grid if necessary, if flex isn't powerful enough to do this without adding scaffolding elements.

brandondong:
There are no issues with centering things horizontally, with or without the wrapping `div`. I added the `div` in order to get the top/bottom vertical whitespace exactly matching with master in all cases. `grid` won't help because it also removes margin collapse. If no wrapping `div` is allowed, then I believe my earlier suggestion of just reducing top margin from 8rem -> 5rem is the best possible approximation of the current whitespace.

acabal:
OK, I think this looks good. Great work, thanks!

PR #203, opened 2022-11-15
Update wanted ebooks list

vr8hub:
I see that sometimes they're deleted and sometimes they're commented out, so I have a 50% of choosing correctly. Of course, I also have a 50% chance of choosing incorrectly…


robinwhittleton:
Typically commented out when someone is actively working on it, and deleted when it’s released 🙂

vr8hub:
Well, then, I was right, which happens far less than 50% of the time. lol (And I wasn't sure, since this one wasn't commented out while I was working on it.)

acabal:
Thanks!

PR #210, opened 2022-12-31
Add ToC to Step-by-step guide

dajare:
- `summary`+`detail` toggle-ToC added before step 1;
- `id` and `title` added to H2 elements;
- CSS rules added to manual.css

Apologies: I thought I was doing this in a branch, but fumbled that, and it's in master. Sorry about that.

### DIFF
--- a/www/contribute/producing-an-ebook-step-by-step.php
+++ b/www/contribute/producing-an-ebook-step-by-step.php
@@ -49,7 +49,7 @@ require_once('Core.php');
 		</details>
 		<ol>
 			<li>
-				<h2 id="setup" title="setup">Set up the Standard Ebooks toolset and make sure it’s up-to-date</h2>
+				<h2 id="setup">Set up the Standard Ebooks toolset and make sure it’s up-to-date</h2>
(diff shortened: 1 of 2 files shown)


acabal:
Thanks! A few notes:

- We don't need to add `title` attributes to the headers. `id` is enough. We also don't need a title on the `<summary>` element.
- We also shouldn't add an ID to the `<detail>` element. You can select it with `.step-by-step-guide detail` instead.

dajare:
Thanks, Alex (and Happy New Year!) -

- I wondered about `title` - that was useful for me in my test, but I see not needed for "production".
- I did try to eliminate the ID on the `detail` and use the CLASS, exactly as you suggest. For whatever reason, I lost the counter function with that arrangement. I'll try again and see if I can work it out.

dajare:
OK - I think those issues are resolved. Not quite sure why I couldn't work out `counter` code before! But this works (hurrah!), and address the tweaks raised above.

acabal:
Great work, thanks!

PR #212, opened 2023-01-11
Capture both digits when rewriting coordinates

dillona:
Without this change only the first digit will be in the rewritten version


acabal:
Thanks!

PR #215, opened 2023-01-15
Basic source folder: No need to edit the title page

colagrosso:
I'm open to guidance on how to word this. I was going to write something like

> Don’t edit this unless the title and author are incorrect (in which case you have bigger problems).

but it was cheeky.


acabal:
Thanks!

Issue #164, opened 2022-02-04
Docker version show blank page


thewchan:
@robinwhittleton It seems that launching the website build via docker stopped working properly; it had worked flawlessly before for me. Does the dockerfile need updating?

robinwhittleton:
I’ve just reinstalled it from scratch and it seems to be working fine. Are you getting any error messages_

thewchan:
Let me try rebuilding the entire containing one more time.

weijia-cheng:
I had the same problem lately, after moving to a new laptop and switched to serving the site locally with Apache.

robinwhittleton:
Did it report any errors? It’s working fine for me, so it’s difficult to debug without knowing what the problem is.

weijia-cheng:
I saw a `PR_END_OF_FILE_ERROR` in Firefox before reloading it a bit and coming up with nothing. It worked fine on my last laptop.

carderne:
I've just cloned the latest version, run `docker build` and when I `docker run` (instructions from the repo) I get a `PR_END_OF_FILE_ERROR` in Firefox, and the following in the Docker logs:

<details closed>
<summary>Click for trace</summary>
<pre>
Generating a RSA private key
..++++
..............................................................++++
writing new private key to '/standardebooks.org/web/config/ssl/standardebooks.test.key'
-----
Loading composer repositories with package information
Warning from https://repo.packagist.org: Support for Composer 1 is deprecated and some packages will not be available. You should upgrade to Composer 2. See https://blog.packagist.com/deprecating-composer-1-support/
Info from https://repo.packagist.org: #StandWithUkraine
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for php-webdriver/webdriver 1.12.1 -> satisfiable by php-webdriver/webdriver[1.12.1].
    - php-webdriver/webdriver 1.12.1 requires ext-curl * -> the requested PHP extension curl is missing from your system.

  To enable extensions, verify that they are enabled in your .ini files:
    - /etc/php/7.4/cli/php.ini
    - /etc/php/7.4/cli/conf.d/10-opcache.ini
    - /etc/php/7.4/cli/conf.d/10-pdo.ini
    - /etc/php/7.4/cli/conf.d/15-xml.ini
    - /etc/php/7.4/cli/conf.d/20-apcu.ini
    - /etc/php/7.4/cli/conf.d/20-calendar.ini
[...]

PR #216, opened 2023-01-16
Fix Docker

robinwhittleton:
We’d had reports of Docker not working (https://github.com/standardebooks/web/issues/164) but I hadn’t been able to replicate. During a new install I ended up with the same problems, which let me debug why I’d previously not seen the problem.

First up, there was a problem with safe, where we needed [a specific newer version](https://github.com/thecodingmachine/safe/releases/tag/v1.3.2) to work with our current setup. Technically this isn’t a required change (`^1.0.0` should be satisfied by a 1.3.2 package) but as it’s also satisfied by lower packages we should fix it.

Secondly, there was a problem with Composer installing packages in the first place. What had happened on my system was that I’d previously installed Composer in my macOS environment and ran `composer install`, so I had a working set of packages in my repo ready for the Docker image to use. This meant that when the image’s `composer install` silently failed due to not being able to contact the outside world, the site carried on working regardless and I never noticed. So the second commit installs the necessary extra `php-` packages and allows Composer to contact the outside world.

This gets my new install back to a working state.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
 		"files": ["lib/Constants.php", "lib/CoreFunctions.php"]
 	},
 	"require": {
-		"thecodingmachine/safe": "^1.0.0",
+		"thecodingmachine/safe": "^1.3.3",
 		"phpmailer/phpmailer": "6.6.0",
 		"ramsey/uuid": "4.2.3",
 		"gregwar/captcha": "1.1.9",
--- a/vms/docker/Dockerfile
+++ b/vms/docker/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:20.04
 
 RUN apt-get update
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y composer php-fpm php-cli php-gd php-xml php-apcu php-mbstring php-intl apache2 apache2-utils libfcgi0ldbl task-spooler libaprutil1-dbd-mysql attr libapache2-mod-xsendfile
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y composer php-fpm php-cli php-gd php-xml php-apcu php-mbstring php-intl php-curl php-zip apache2 apache2-utils libfcgi0ldbl task-spooler libaprutil1-dbd-mysql attr libapache2-mod-xsendfile
 RUN DEBIAN_FRONTEND=noninteractive apt-get install -y sudo imagemagick openjdk-8-jre python3 pip calibre
 RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 


vms/docker/Dockerfile
```diff
@@ -19,5 +19,8 @@ RUN a2enmod headers expires ssl rewrite proxy proxy_fcgi authn_dbd authn_socache
 # Disable opcaching for dynamic PHP reloading
 RUN echo "opcache.enable=0" >> /etc/php/7.4/fpm/php.ini
 
+# Allow composer to download packages
```
acabal, line 22: I don't think we want to be appending directly to the ini file, because in the setup instructions in the readme it's symlinked to the copy in the repo. So if you were to update it here, the repo would appear dirty because we've worked on a symlink.

acabal: Also, we specifically disable allow_url_fopen in our current ini file for security reasons. While it's not a *huge* deal to enable it in a docker container, I don't think hot-altering the site's configuration just to get composer to work is the right direction. If you nee to alter it, it should happen in a docker-specific PHP config file and not the site's own config file.

composer.json
```diff
@@ -8,7 +8,7 @@
 		"files": ["lib/Constants.php", "lib/CoreFunctions.php"]
 	},
 	"require": {
-		"thecodingmachine/safe": "^1.0.0",
+		"thecodingmachine/safe": "^1.3.3",
```
acabal, line 11: I don't understand why upgrading Safe is required to get Docker working?

robinwhittleton: It’s not. What we need is a minimum installed package level of safe 1.3.2 to get the site running, due to line https://github.com/standardebooks/web/blob/ebf9651e25b17788a699bc6d09b8613a68b6742d/lib/Constants.php#L4. As it stands, specifying `^1.0.0` will install 1.3.3 due to version resolution, but my usual feeling is that if we need a specific minimum package version we might as well require it.

So what I’m saying is that I’m happy to drop this commit if you’d rather 🙂

robinwhittleton: I just checked and we’re running safe 1.3.3 on the server.

acabal: OK!

vms/docker/Dockerfile
robinwhittleton: OK, I’m a little confused at this point. It seems like we don’t need the second commit at all, because `composer install` happens before the symlink that shuts off outside access. I’d been testing this by manually running it from the terminal, but of course at that point the access was locked down.

So what we end up as being my problem was just an out of date vendor folder. Which means that I still have no idea why everyone else was having Docker problems.

robinwhittleton: Ah, but of course we still need php-curl and php-zip. So I’ll amend the PR to just be that and the safe update.

acabal:
OK thanks! I suppose we'll just have to keep an eye on it to see if people still have problems. Unfortunately I still have zero knowledge on how to use Docker. Or, maybe you could ask in the mailing list to see if someone can help figure out what's going on.

PR #217, opened 2023-02-04
Update Wanted Guardian list link from 2003 to 2015

dajare:
The link in the "Wanted" list goes to the (outdated) 2003 version, not the 2015 list which is the [SE Collection][1] version.

[1]: https://standardebooks.org/collections/the-guardians-best-100-novels-in-english-2015


acabal:
Thanks!

PR #221, opened 2023-02-16
Don't escape forward slash in grep expression

bentley:
Fixes a warning with GNU Grep 3.8:

```
grep: warning: stray \ before /
```


acabal:
Thanks!

Issue #220, opened 2023-02-16
Duplicate repo causes sync errors


bentley:
There’s a fork of [dorothy-day_the-eleventh-virgin](https://github.com/standardebooks/dorothy-day_the-eleventh-virgin) at [dorothy-day_the-eleventh-virgin-1](https://github.com/standardebooks/dorothy-day_the-eleventh-virgin-1), and of [voltairine-de-cleyre_short-fiction](https://github.com/standardebooks/voltairine-de-cleyre_short-fiction) at [voltairine-de-cleyre_short-fiction-1](https://github.com/standardebooks/voltairine-de-cleyre_short-fiction-1).

This causes sync-ebooks to fail: if you run it from the same directory multiple times periodically (as I do), the script renames the extra repository from *repo*-1.git to *repo*.git. That executes correctly if *repo*.git doesn’t exist—but if it does exist, it executes incorrectly, but still without failure, the first time, moving *repo*-1.git to *repo*.git/*repo*-1.git and then cloning it again:

```
Moving dorothy-day_the-eleventh-virgin-1.git to dorothy-day_the-eleventh-virgin.git
Cloning https://github.com/standardebooks/voltairine-de-cleyre_short-fiction-1.git ... 
Done.              
```

The next time sync-ebooks is run, it fails, because *repo*.git/*repo*-1.git already exists, and *this* failure causes the whole script to exit:

```
Moving voltairine-de-cleyre_short-fiction-1.git to voltairine-de-cleyre_short-fiction.git
mv: cannot move 'voltairine-de-cleyre_short-fiction-1.git' to 'voltairine-de-cleyre_short-fiction.git/voltairine-de-cleyre_short-fiction-1.git': Directory not empty
```

The same happens with guy-de-maupassant_short-fiction_ernest-boyd_storm-jameson_jeffery-e-jeffery_lafcadio-hearn_m-walter-.git, which gets renamed to guy-de-maupassant_short-fiction_ernest-boyd_storm-jameson_jeffery-e-jeffery_lafcadio-hearn_m-walter-dunne_henry-c-olinger_albert-m-cohn-mcmaster_dora-knowlton-ranous_bigelow-brown-co-inc_francis-steegmuller.git. (It doesn’t happen with other so‐long‐they’re‐truncated repo names, as far as I can tell.)

To run sync-ebooks repeatedly in the same directory, I consistently have to remove these three repositories first.

acabal:
[...]

PR #222, opened 2023-02-17
sync-ebooks: Don't rename repository if target directory already exists

bentley:
The rename code doesn’t behave as intended if the directory already exists, silently moving the old directory into the new directory rather than renaming it. Handling this gives a visible diagnostic instead of invisibly misbehaving, and also allows the script to continue without terminating early.


bentley:
I also figured out why I was getting duplicates of repos with very long names in #220: the script attempts to lengthen truncated names with `${repoName}*/`, but if `-b` is passed, doesn’t account for the trailing `.git`.

acabal:
Thanks!

PR #224, opened 2023-03-01
update wanted ebooks list (2 released books)

ConcaveTrillion:
2 released books are still on the list, remove them


acabal:
Thanks!

PR #225, opened 2023-03-09
Two minor process improvements to the step-by-step guides

weijia-cheng:
1. In the producer guide, provide the `xpath` command from the reviewer guide to find instances of all-caps instead of the current regex. Not only is the xpath command easier to use (just paste it in), it covers more edge cases, so I don't see why producers shouldn't use it too.
2. In the reviewer guide, add a reminder to `lint` at the start of the review. I figure most reviewers do this anyways, but it's good to have an explicit reminder so you don't forget. Otherwise you'll get all the way to the bottom before the guide tells you to lint, if you forgot to do that at the start.


acabal:
Good additions, thanks!

Issue #126, opened 2021-05-08
Primitive Browser Support


npars:
The Standard Ebooks website doesn't render correctly in some E-Reader browsers. For instance the Kobo Libra H20: 
![PXL_20210508_171029248](https://user-images.githubusercontent.com/1280422/117547777-dd5c3780-b007-11eb-990a-56e2678e0b31.jpg)

Since the Kobo can download ebooks directly through the browser it would be great if the site was easier to navigate. Would it be possible to create a static "lite" version of the site, similar to the mobile websites that used to be common?

acabal:
Maybe a good long term goal if someone is interested in working on it. We could serve a different stylesheet to browsers matching the Kobo user agent. But, debugging this would be extremely tedious.

robinwhittleton:
I spent a while seeing if I could match the Kobo browser with specific media queries, but it looks like it just thinks it’s a normal browser on a normal machine. It doesn’t respond to `@media (monochrome)` or `@media (max-color: 4)` in any different way that my laptop. Unfortunate.

npars:
> I spent a while seeing if I could match the Kobo browser with specific media queries, but it looks like it just thinks it’s a normal browser on a normal machine. It doesn’t respond to `@media (monochrome)` or `@media (max-color: 4)` in any different way that my laptop. Unfortunate.

Could we key off of the user agent? My Kobo returns: 
```
Mozilla/5.0 (Linux; U; Android 2.0; en-us;) AppleWebKit/538.1 (KHTML, like Gecko) Version/4.0 Mobile Safari/538.1 (Kobo Touch 0384/4.30.18838)
```

acabal:
User agent is an option. Check it in php and output a different stylesheet

robinwhittleton:
How stripped back do we want the ereader experience to be? My default inclination is pretty stripped back, but keeping the logo and fonts as part of our branding.

I’m currently building a completely separate ereader.css file that doesn’t include anything else. Once I’ve got something good working I’ll work out what’s shared and maybe make a base.css that both can include.

acabal:
Yeah I'm not certain. As much as we can get away with, I suppose. If we're targeting eink then the big factor that will influence most of it is going to be what the browser supports, then color, and then the poor tap responsiveness of eink. I assume Kobo is just using Webkit. I don't know what Kindle uses but I'm sure it's awful.

robinwhittleton:
[...]

PR #226, opened 2023-03-19
Primitive browser support

colagrosso:
See discussion in issue #126


acabal:
This looks great! The only comment I would make is that the `EreaderBrowser` class isn't really necessary, as it's only used in one place and is not a logical "object" per se. Instead we can just put that `if` logic straight into the header template. If later on we find that we need to make that same check somewhere else, we can create a class or reusable function then.

colagrosso:
Thanks for the quick review. We need the check in `Header.php` and `SearchForm.php`. I'm open to a different way to make a reusable function because you're right that it's not a logical object.

colagrosso:
Good call, done.

acabal:
Excellent, thanks!

PR #228, opened 2023-04-02
Remove PollItem property from Session

colagrosso:
I think this was a copy–paste error when the file was created in 0bc3dc3 with Session.php based on PollVote.php.


acabal:
Thanks!

Issue #234, opened 2023-06-06
Cover art database


acabal:
Also see https://github.com/standardebooks/web/issues/232 and https://github.com/standardebooks/web/pull/233

We need a cover art database that can do the following:

- List approved paintings that can be used for covers, browsable/searchable by:
    - Keyword like `castle` or `seascape`
    - Artist name
    - Painting name
- Detail page for each artwork listing artist, dates, hi-res download link, whether the cover is in use in the corpus
- Form for anonymous contributors to submit artworks for approval
    - Submissions can be anonymous
    - Captcha to prevent bots - see the newsletter signup form for a basic captcha implementation
    - Submissions enter a queue to be approved by a moderator before going live on the site
- Moderator admin page
    - List of artworks up for approval
    - Yes/no button to approve an artwork

Out of scope, to be explored at a later date:
- User login/management system - for now submissions will be anonymous without requiring an account or login; for now we can simply password-protect the mod admin page via Apache
- Editing metadata of artwork that has already been submitted
- Any other mod function besides approving/declining artwork in the queue

## Artwork requirements

- Artist name - we should store artists with their info in a table so we don't duplicate all that information

- Artist death year

- Artwork name

- Artwork completed year (may be circa, a range, or unknown as well)

- Color image upload

- A list of subject tags for the image, like "castle" or "seascape"

- PD proof:

    - Link to an approved museum page 
 
    OR ALL OF THE FOLLOWING

    - Year book was published
    - Link to direct page scan of artwork (not just the start of the book, the direct page)
    - Link to direct page scan of page mentioning book publication year (not just the start of the book, the direct page)
    - Link to direct page scan of book copyright/rights statement page (may be the same as the publication year page but not always)

## DB schema

Use the `se` database on `localhost`

Tables:

- Artworks

- Artists

- ArtworkTags

- Users - SE already has a Users table we can reuse, or we may skip this if we just password-protect the mod page

## URL schema

Use a REST-like schema for our URLs.

See the [polls system on the SE website](https://github.com/standardebooks/web/tree/master/www/polls) for a model to follow for a RESTful service with forms for users to submit data.

- `GET <root>/artworks/new` -- artwork submission form
- `POST <root>/artworks` -- create new artwork
- `GET <root>/artworks` -- browse the list of approved artwork
- `GET <root>/artworks/<artist-slug>/<artwork-slug>` -- view a specific artwork entry
- `GET <root>/admin/artworks` -- view the queue of unapproved artwork
[...]

PR #235, opened 2023-06-10
Initial cover art database schema and Artwork and Artist classes

colagrosso:
Progress on #234

### DIFF
--- a/config/sql/se/Artists.sql
+++ b/config/sql/se/Artists.sql
@@ -1,0 +1,7 @@
+CREATE TABLE `Artists` (
+  `ArtistId` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `Name` varchar(191) NOT NULL,
+  `DeathYear` smallint unsigned NULL,
+  PRIMARY KEY (`ArtistId`),
+  UNIQUE KEY `idxUnique` (`Name`,`DeathYear`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/config/sql/se/ArtworkTags.sql
+++ b/config/sql/se/ArtworkTags.sql
@@ -1,0 +1,5 @@
+CREATE TABLE `ArtworkTags` (
+  `ArtworkId` int(11) unsigned NOT NULL,
+  `TagId` int(10) unsigned NOT NULL,
+  PRIMARY KEY `idxUnique` (`ArtworkId`,`TagId`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/config/sql/se/Artworks.sql
+++ b/config/sql/se/Artworks.sql
@@ -1,0 +1,12 @@
+CREATE TABLE `Artworks` (
+  `ArtworkId` int(11) unsigned NOT NULL AUTO_INCREMENT,
+  `ArtistId` int(10) unsigned NOT NULL,
+  `Name` varchar(255) NOT NULL,
+  `UrlName` varchar(255) NOT NULL,
+  `CompletedYear` smallint unsigned NULL,
+  `CompletedYearIsCirca` boolean NOT NULL DEFAULT FALSE,
+  `ImageFilesystemPath` varchar(255) NOT NULL,
+  `Created` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  `Status` enum('unverified', 'approved', 'declined', 'in_use') DEFAULT 'unverified',
+  PRIMARY KEY (`ArtworkId`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/config/sql/se/Tags.sql
+++ b/config/sql/se/Tags.sql
@@ -1,0 +1,7 @@
+CREATE TABLE `Tags` (
+  `TagId` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `Name` varchar(80) NOT NULL,
+  `UrlName` varchar(80) NOT NULL,
+  PRIMARY KEY (`TagId`),
+  UNIQUE KEY `idxUnique` (`UrlName`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/lib/Artist.php
+++ b/lib/Artist.php
@@ -1,0 +1,43 @@
+<?
+use Safe\DateTime;
+
+class Artist extends PropertiesBase{
+	public $ArtistId;
+	public $Name;
+	public $DeathYear;
+
+	// *******
+	// METHODS
+	// *******
+
+	protected function Validate(): void{
+		$error = new Exceptions\ValidationException();
+
+		if($this->Name === null || strlen($this->Name) === 0){
+			$error->Add(new Exceptions\InvalidArtistException());
+		}
+
+		if($error->HasExceptions){
+			throw $error;
+		}
+	}
+
+	public function Create(): void{
+		$this->Validate();
+		Db::Query('
+			INSERT into Artists (Name, DeathYear)
+			values (?,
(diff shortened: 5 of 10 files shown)


config/sql/se/ArtworkTags.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE `ArtworkTags` (
+  `ArtworkId` int(11) unsigned NOT NULL,
+  `SubjectTag` varchar(80) NOT NULL,
+  UNIQUE KEY `idxUnique` (`ArtworkId`,`SubjectTag`)
```
jobcurtis, line 4: Tiny suggestion but this unique index might as well be a primary key

```suggestion
  PRIMARY KEY (`ArtworkId`,`SubjectTag`)
```

config/sql/se/Artists.sql
jobcurtis, line 1: I think we may want some kind of unique constraint, either on just the name or on Name + DeathYear. When the form is submitted we won't have the artist ID so we'll have to select based on the information available in the form.

I guess it's possible that there could be two artists with the same name who died in the same year, but we'd have no way to differentiate between them anyway so they might as well be one record.

config/sql/se/Artists.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE `Artists` (
+  `ArtistId` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `Name` varchar(255) NOT NULL,
+  `DeathYear` datetime DEFAULT NULL,
```
jobcurtis, line 4: Is there much value in storing this as a `datetime` rather than `int`? We're only asking for the year, so the month, day & time portions will all be unused.

acabal:
Great, this looks really good. You have a good very grasp of the code style and patterns we're using!

Some notes:

- I think Job is right, artist death year can just be an `UNSIGNED SMALLINT|NULL` with `null` meaning "unknown".
- For artworks, completed year can be unknown, circa, or an integer. Storing it all as a string might not be sufficient, because we want the option to later implement a sort by completed year, or filter by completed year, etc. So we need to think of a different schema for completed year - maybe `CompletedYear - UNSIGNED SMALLINT|NULL` and `CompletedYearIsKnown - BOOL DEFAULT TRUE`? Then, if `CompletedYearIsKnown == true`, `CompletedYear` is the actual year. If `CompletedYearIsKnown == false`, an integer value for `CompletedYear` is interpreted as "circa that year", and a null value is interpreted as "unknown". Thoughts?
- `Artworks.UrlName` in the DB should not include the artist name, because if we change the artist name then we have to update all of the artworks in the DB at the same time. Instead, `Artworks.UrlName` should be just the artwork name. Then, when we want to get the *complete* UrlName in PHP, we use the getter `Artwork::GetUrlName()` which constructs it by getting the `Artwork::Artist`. To see this in action see `PollVote.php:22`, which fires when we do for example `print($pollVote->Url)`.
- `ProperitesBase` allows you to access a property within the class without using a leading `_`. See for example `NewsletterSubscription.php:23`. The benefit of doing it this way is that the getter is called instead, which may prefill a property; and the code looks more consistent because we never end up referencing private/protected properties.
- I would suggest not using a big constructor for these objects. In my experience it's clearer to read when we simply create a blank object and assign properties individually, especially since we'll be using `HttpInput` to get request data.

These are all just nitpicks and on the whole I think we're most definitely on the right track here. Great work!

acabal:
I've given both of you write access to the repo so you should be able to merge PRs as necessary. I don't really have a vision for how this should be managed; you both can push as you see fit, or ask for review or feedback before you do so. Up to you guys.

acabal:
Maybe for artwork completed year, the field should be called `CompletedYearIsCirca`. Then, if it's true, the year is interpreted as circa. If the completed year is null, then that's interpreted as "unknown" regardless of the value of `CompletedYearIsCirca`. Maybe that naming is clearer than what I originally proposed?

acabal:
Also, `ArtworkTags` should map an artwork ID to a tag ID. Therefore, ArtworkTags should look like:

- ArtworkId: int
- TagId: int

and we also need a `Tags` table that looks something like:

- TagId: int primary key
- Name: varchar
- UrlName: varchar

This is a more normalized schema and will give us flexibility in editing tag names/URLs in the future.

config/sql/se/ArtworkTags.sql
colagrosso: Sure, done.

config/sql/se/Artists.sql
colagrosso: Good call, done. I added the unique constraint on Name + DeathYear.

I didn't add a method to `Artist.php` to select an existing artist, but I'll do that in a followup PR.

colagrosso:
Thanks for all the comments, Job and Alex! 

A few responses inline to Alex:

> * I think Job is right, artist death year can just be an `UNSIGNED SMALLINT|NULL` with `null` meaning "unknown".

Yep, done.

> * For artworks, completed year can be unknown, circa, or an integer. Storing it all as a string might not be sufficient, because we want the option to later implement a sort by completed year, or filter by completed year, etc. So we need to think of a different schema for completed year - maybe `CompletedYear - UNSIGNED SMALLINT|NULL` and `CompletedYearIsKnown - BOOL DEFAULT TRUE`? Then, if `CompletedYearIsKnown == true`, `CompletedYear` is the actual year. If `CompletedYearIsKnown == false`, an integer value for `CompletedYear` is interpreted as "circa that year", and a null value is interpreted as "unknown". Thoughts?

I went with a new field called `CompletedYearIsCirca`, as you had in your followup suggestion. Seems fine unless we want to support artwork like this:

https://standardebooks.org/ebooks/h-g-wells/the-war-of-the-worlds/text/colophon

which has "a painting completed in the 17th century."


> * `Artworks.UrlName` in the DB should not include the artist name, because if we change the artist name then we have to update all of the artworks in the DB at the same time. Instead, `Artworks.UrlName` should be just the artwork name. Then, when we want to get the _complete_ object URL in PHP, we use the getter `Artwork::GetUrl()` which constructs it by getting the `Artwork::Artist`. To see this in action see `PollVote.php:22`, which fires when we do for example `print($pollVote->Url)`. This method makes heavy use of the `PropertiesBase` syntactic sugar.

Got it, thanks for the `PollVote.php:22` example. I haven't got to this step yet, but when it's time to implement this URL:

We'll have to store an `Artist.UrlName` too, which we don't have yet. And then implement something like:

```
public static function Get(?string $artistUrlName, ?string $artworkUrlName): Artwork{}
```

> * `ProperitesBase` allows you to access a class's property _within the class itself_ without using a leading `_`. See for example `NewsletterSubscription.php:23`. The benefit of doing it this way is that the getter is called instead, which may prefill a property; and the code looks more consistent because we never end up referencing private/protected properties.

Got it, thanks. I removed the superfluous underscores. There is one that I left in `Artwork.Create()` because I didn't want the getter to trigger a database lookup. 

> * I would suggest not using a big constructor for these objects. In my experience it's clearer to read when we simply create a blank object and assign properties individually, especially since we'll be using `HttpInput` to get request data.

Sure, done.

colagrosso:
> I've given both of you write access to the repo so you should be able to merge PRs as necessary. I don't really have a vision for how this should be managed; you both can push as you see fit, or ask for review or feedback before you do so. Up to you guys.

Thanks for that. 

@jobcurtis, here's what I propose for managing this branch while we are just starting out:

* We ask each other for a review and wait for an "LGTM" before merging
* You can merge typos or other trivial fixes without review

If this slows us down too much, we can change.

acabal:
Sounds good!

jobcurtis:
Yeah I think that sounds reasonable. This PR looks good to me.

PR #236, opened 2023-06-12
Initial artwork submission form

jobcurtis:
#234

Here's a draft PR for the artwork submission form. It's branched from #235 so includes commits from that PR as well (that PR should be merged before this one). It's not complete yet. The main things missing still are;

- [x] validation / storage of PD proof fields
- [x] allow specifying that artwork completed year is approximate
- [x] creation of thumbnails for uploads
- [ ] general form UX - we definitely need some help text / placeholder values to guide users a bit, and the layout could probably be improved.


jobcurtis:
Hey @colagrosso, I think this is ready for a second pair of eyes. This PR contains;

- required fields for submitting a new artwork
- basic validation
- checking that uploaded images are actually images, storing them & creating a thumbnail

Things that aren't in this PR but should be or could be included later;

- more helpful feedback when failing validation
- previously filled-in fields should be set when you're redirected to the form after hitting an error
- more validation around the pd-proof fields, particularly the museum URL option
- guidance text / placeholder values

Also note that apache is currently configured to reject requests where the body is > 2MB -- I think we will need to add an exception to that rule for the `/artworks/post` route.

www/artworks/post.php
```diff
@@ -0,0 +1,139 @@
...
+
+	if ($uploadInfo[2] !== IMAGETYPE_GIF && $uploadInfo[2] !== IMAGETYPE_JPEG && $uploadInfo[2] !== IMAGETYPE_PNG) {
+		throw new Exceptions\InvalidImageUploadException();
+	}
+
+	$uid = Uuid::uuid4()->toString();
+	$ext = image_type_to_extension($uploadInfo[2]);
+
+	$uploadPath = COVER_ART_UPLOAD_PATH . $uid . $ext;
+	$thumbPath = COVER_ART_UPLOAD_PATH . $uid . '.thumb' . $ext;
```
colagrosso, line 26: It's worth taking a moment to ensure we like this naming convention. We end up with files that look like this:

```
$ ls -1 www/images/uploads/
02584f01-fc89-4dec-aeff-9db5f8b97055.jpeg
02584f01-fc89-4dec-aeff-9db5f8b97055.thumb.jpeg
0ee63a18-eda7-4289-9a95-d5e6742ec08e.jpeg
0ee63a18-eda7-4289-9a95-d5e6742ec08e.thumb.jpeg
361a0483-940f-4520-b09a-d36f89c937c0.jpeg
361a0483-940f-4520-b09a-d36f89c937c0.thumb.jpeg
97063843-0b29-48d8-a215-14898fbf1dff.jpeg
97063843-0b29-48d8-a215-14898fbf1dff.thumb.jpeg
c3803beb-a3b6-4e18-977b-8c3ab6b9bec8.jpeg
c3803beb-a3b6-4e18-977b-8c3ab6b9bec8.thumb.jpeg
cfa67358-40eb-44d2-abef-5b67bbea1e77.jpeg
cfa67358-40eb-44d2-abef-5b67bbea1e77.thumb.jpeg
ea2b1c23-df11-49ef-bdc7-00d0a17ca353.jpeg
ea2b1c23-df11-49ef-bdc7-00d0a17ca353.thumb.jpeg```

Alternatives

1. A `thumbs/` subdirectory
2. A different separator, e.g., `_thumb.jpeg

The reason I ask is that we'll need to add a new property to the `Artwork` class:

```
_ThumbnailUrl
```

that is derived from `ImageFilesystemPath`. We might want to make it easy on ourselves to derive the thumbnail URLs. Maybe the dot will work, but we'll have to careful with the string or regex logic.

File: lib/Constants.php [hunk 2]
```diff
@@ -34,6 +35,8 @@
 const SORT_READING_EASE = 'reading-ease';
 const SORT_LENGTH = 'length';
 
+const COVER_THUMBNAIL_SIZE = 300;
```
colagrosso, line 38: I think we could bump this up to 350.

www/artworks/post.php
```diff
@@ -0,0 +1,139 @@
...
+
+	if ($src_h > $src_w) {
+		$dst_h = COVER_THUMBNAIL_SIZE;
+		$dst_w = intval($dst_h * ($src_w / $src_h));
+	} else {
+		$dst_w = COVER_THUMBNAIL_SIZE;
+		$dst_h = intval($dst_w * ($src_h / $src_w));
+	}
+
+	switch ($uploadInfo[2]) {
```
colagrosso, line 43: I think we should accept only JPEG uploads. It would simplify this code. Plus, the eventual ebook producer will need to supply a `cover.source.jpg`, so a PNG or GIF will be less useful.

www/artworks/post.php
```diff
@@ -0,0 +1,139 @@
...
+	$artwork = new Artwork();
+
+	$imageUpload = handleImageUpload($_FILES['color-upload']['tmp_name']);
+
+	$artwork->Artist = $artist;
+	$artwork->Name = HttpInput::Str(POST, 'artwork-name', false);
+	$artwork->CompletedYear = HttpInput::Str(POST, 'artwork-year', false);
+	$artwork->CompletedYearIsCirca = HttpInput::Bool(POST, 'artwork-year-is-circa', false);
+	$artwork->ArtworkTags = parseArtworkTags();
+	$artwork->ImageFilesystemPath = $imageUpload[0];
```
colagrosso, line 112: This leads to database fields that look like this:

```
/standardebooks.org/web/www/images/uploads/c01d7ac5-09e2-4e32-ab3f-985dab0f7d2b.jpeg
```

which I think is not quite right. Let's not write the `WEB_ROOT` to the database, just something like this:

```
/images/uploads/c01d7ac5-09e2-4e32-ab3f-985dab0f7d2b.jpeg
```

That way, it will also serve as a download URL.

www/artworks/new.php
```diff
@@ -0,0 +1,144 @@
...
+						<input type="number" id="artwork-year" name="artwork-year" min="0" max="<?= gmdate('Y') ?>"/>
+					</label>
+				</div>
+				<label>
+					Artwork Tags
+					<input type="text" name="artwork-tags" placeholder="tags, comma-separated"/>
+				</label>
+			</fieldset>
+			<fieldset id="pd-proof">
+				<legend>Proof of Public Domain Status</legend>
```
colagrosso, line 84: I think you noted this, too, but we should give some guidance to the form submitter here. Maybe as a short paragraph, maybe as some tooltips. (We can do this in a subsequent PR.)

README.md
```diff
@@ -15,6 +15,10 @@ sudo mkdir /var/log/local/
 cd /standardebooks.org/
 git clone https://github.com/standardebooks/web/
 
+# Create & grant ownership of image uploads directory
+mkdir /standardebooks.org/web/www/images/uploads/
```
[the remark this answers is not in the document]
jobcurtis: A `.gitignore` that ignores everything in the directory (except itself) with an explanatory comment should do the trick. Will add.

www/artworks/new.php
```diff
@@ -0,0 +1,144 @@
...
+			</p>
+		<? } ?>
+
+		<form method="post" action="/artworks" enctype="multipart/form-data">
+			<fieldset>
+				<legend>Artist Details</legend>
+				<div>
+					<label>
+						Artist Name
+						<datalist id="artist-names">
```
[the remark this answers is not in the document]
jobcurtis: I agree that it needs refinement - what if we eventually have works by 10,000 artists, are we going to include them all in the form html? Hopefully not.

I suppose the most straightforward way to achieve autofill for the death year field would be an accompanying JS script that retrieves that data from a `data-death-year` attribute on the artist in the datalist. We would want to get Alex's OK on that first though as there is no JS elsewhere in the site. In this case it might be alright as it wouldn't be providing critical functionality -- users with scripts disabled would just be missing autofill for one field, which isn't the end of the world.

lib/Artwork.php
```diff
@@ -85,10 +92,6 @@ protected function Validate(): void{
 			$error->Add(new Exceptions\InvalidArtworkException());
 		}
 
-		if($this->CompletedYear === null || strlen($this->CompletedYear) === 0){
```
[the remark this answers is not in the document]
colagrosso: Ah, sorry, you're right of course. My mistake.

www/artworks/new.php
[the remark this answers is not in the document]
colagrosso: It's not an issue of scale, just convenience and correctness. 

Let me think about this some more and maybe prototype something. I was imagining a form where the user would either pick an existing artist (both name and death year), or enter both manually. What we have now is a mix where the user picks the name but must always enter the death year manually.

No Javascript is a [core principle](https://alexcabal.com/posts/standard-ebooks-and-classic-web-tech), so I would prefer not to add it here.

www/artworks/post.php
acabal: GUIDs are not very ergonomic in the case where we have to go in and examine or manipulate files by hand. Instead we can simply name the file after the DB primary key value (i.e. `1.jpg`, `2.jpg`, etc.). That way we also avoid needing another column for GUID. Also let's go with `.jpg` and not `.jpeg`.

acabal:
Let me expand on what I wrote about JS a little because this fuzzy-match textbox is a good example.

We should design the system to work without JS in any case, because the user may have JS disabled and they should still be able to use the system. However, it's clear that often JS enhances a UI. So we can add JS later as an *enhancement*, but the underlying system should still work without it.

The datalist for artist names is a good example of this. `<datalist>` may not be good enough to fuzzy match alternate spellings and accent marks. So the system should probably prompt the user after POST to disambiguate those issues, before actually recording the form data:

1. User enters `Ilia Repyn` (an alternate spelling) into the form and submits
2. POST function checks the DB for variant spellings and sees the user probably meant `Ilya Repin`. 
3. Instead of showing confirmation page, user is returned to the form with a messages saying `Did you mean Ilya Repin?`
4. User can select the alternate spelling we suggest, or submit using original name

This is a flow without JS, but that would be clearly enhanced with a fully-featured fuzzy match JS textbox, because we could easily disambiguate before the form is submitted. So we can explore adding such a fuzzy JS textbox later, but the important point is that the flow works without JS anyway.

www/artworks/post.php
jobcurtis: If the filename is going derived from the PK we can remove the column that stores the file location, right? As the file location will always be `/www/images/uploads/{id}.jpg`.

README.md
```diff
@@ -15,6 +15,10 @@ sudo mkdir /var/log/local/
...
+sudo chown www-data:www-data /standardebooks.org/web/www/images/uploads/
```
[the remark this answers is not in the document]
jobcurtis: This was a guess on my part as I'm not running Ubuntu so my apache setup is a bit different. I'll remove this and make a note on the issue that we need to figure out the correct steps for the readme.

www/artworks/new.php
jobcurtis: Agreed, let's leave this for later once the functionality of the form is complete.

lib/Artist.php
```diff
@@ -40,4 +40,32 @@ public function Delete(): void{
 			where ArtistId = ?
 		', [$this->ArtistId]);
 	}
+
+    public static function GetOrCreate(string $name, ?int $deathYear): Artist {
+        $result = Db::Query('
+            SELECT ArtistId, Name, DeathYear
```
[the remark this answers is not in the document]
jobcurtis: I'd prefer to be explicit about what's being selected (and in what order) unless there's a good reason not to. These tables aren't necessarily set in stone and `SELECT *` is prone to quietly break if the table definition changes.

www/artworks/post.php
```diff
@@ -0,0 +1,97 @@
...
+
+	return $uploadPath;
+}
+
+/**
+ * @return array<ArtworkTag>
+ * @throws \Exceptions\InvalidArtworkTagException
+ */
+function parseArtworkTags(): array {
+	$artworkTags = HttpInput::Str(POST, 'artwork-tags', false);
```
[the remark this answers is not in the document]
jobcurtis: Good catch, I've fixed this (and the issue with the empty tags)

lib/Artist.php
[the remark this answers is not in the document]
colagrosso: I understand your reasoning, but I think consistency with the codebase is just as important. There's a pretty consistent use of `SELECT *`, including for PropertiesBase objects: 

https://github.com/search?q=repo%3Astandardebooks%2Fweb%20%22select%20*%22&type=code

www/artworks/post.php
colagrosso: @acabal, how do feel about the full-resolution image URLs being guessable? Going with `1.jpg`, `2.jpg`, etc. would allow potentially annoying scrapers:

```
for i in $(seq 1 500); do
  curl -O https://standardebooks.org/www/images/uploads/$i.jpg
done
```

(This came up in a [thread](https://groups.google.com/g/standardebooks/c/KrXyzZVZig4/m/E7BRuJ2wBQAJ) in Mar–Apr 2021 about bulk downloading ebooks.)

If you don't think that's too big of an issue, or it's still worth the trade-off, then I agree with you and Job that we can simplify here and remove the `ImageFilesystemPath` column. We'll still add these to properties to the `Artwork` class,

```
_Url
_ThumbnailUrl
```

but they will be trivial.

acabal: Having them guessable is fine. Unlikely to be scraped I think, nobody is seriously scraping hi-res public domain oil paintings without any accompanying metadata. If it becomes a problem we can address it at that time.

www/artworks/post.php
jobcurtis: Good point, I've done that.

www/artworks/post.php
jobcurtis: Good catch, I've fixed this so that `GetImageUrl()` returns e.g. `/images/uploads/1.jpg`.

www/artworks/post.php
jobcurtis: Cool, I've implemented this change so that uploads are stored at `/images/uploads/{id}.jpg` and thumbs at `/images/uploads/{id}.thumb.jpg`.

Another potential risk of is that if upload URLs are predictable, it would be possible to use the SE website as a free image hosting service since even unapproved uploads will be available. This also seems pretty unlikely (and not too difficult to detect) but maybe worth being aware of.

www/artworks/new.php
```diff
@@ -0,0 +1,144 @@
...
+		<hgroup>
+			<h1>Submit Artwork</h1>
+			<h2></h2>
+		</hgroup>
+
+		<?= Template::Error(['exception' => $exception]) ?>
+
+		<? if ($success) { ?>
+			<p class="message success">
+				Thank you for submitting artwork. It will be reviewed and added to the database if it is approved.
```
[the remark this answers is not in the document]
jobcurtis: Good idea, I've added the message with the name of the artwork in it.

www/artworks/new.php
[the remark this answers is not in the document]
jobcurtis: Cool - let's make a note on the main issue to handle this better at some point down the road.

www/artworks/post.php
```diff
@@ -0,0 +1,139 @@
+<? /** @noinspection PhpComposerExtensionStubsInspection, PhpIncludeInspection */
+
+use Ramsey\Uuid\Uuid;
+
+require_once('Core.php');
+
+/**
+ * @return array<string> an array containing [0] the path to the uploaded image and [1] the path to the thumbnail
```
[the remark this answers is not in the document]
jobcurtis: This function now returns `void` as the image/thumb paths are generated by the `Artwork` class instead.

jobcurtis:
Alright, I think I've addressed all the initial feedback and it's ready for another look (@colagrosso). I'll handle the merge conflict shortly.

Regarding the autofill/search for existing artists, I think let's leave that for a separate PR after we've decided how it should be implemented. For this PR I'm happy to leave in the current datalist or take it out, whatever we think is appropriate.

www/artworks/post.php
```diff
@@ -0,0 +1,122 @@
...
+ */
+function handleImageUpload(string $uploadTmp, Artwork $artwork): void{
+	$uploadInfo = getimagesize($uploadTmp);
+
+	if ($uploadInfo === false){
+		throw new Exceptions\InvalidImageUploadException();
+	}
+
+	if ($uploadInfo[2] !== IMAGETYPE_JPEG){
+		throw new Exceptions\InvalidImageUploadException("Uploaded image must be a JPG file.");
```
colagrosso, line 16: nit: single quotes

www/artworks/post.php
```diff
@@ -0,0 +1,122 @@
...
+	$artwork->PublicationYear = HttpInput::Int(POST, 'pd-proof-year-of-publication');
+	$artwork->PublicationYearPage = HttpInput::Str(POST, 'pd-proof-year-of-publication-page', false);
+	$artwork->CopyrightPage = HttpInput::Str(POST, 'pd-proof-copyright-page', false);
+	$artwork->ArtworkPage = HttpInput::Str(POST, 'pd-proof-artwork-page', false);
+
+	$artwork->Create();
+
+	handleImageUpload($_FILES['color-upload']['tmp_name'], $artwork);
+
+	$_SESSION['success-message'] = "“" . $artwork->Name . "” submitted successfully!";
```
colagrosso, line 104: nit: single quotes

www/artworks/post.php
```diff
@@ -0,0 +1,122 @@
...
+	http_response_code(303);
+	header('Location: ' . "/artworks/new");
```
colagrosso, line 106: nit: single quotes

www/images/uploads/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Ignore everything
+*
+
+# Except this file
+!.gitignore
+
+# This directory (/www/images/uploads) contains image uploads and thumbnails that
+# should not be checked into git.
```
colagrosso, line 8: nit: I think this red circle is indicating that there's not a newline at the end of this file. If that's the case, please add one.

colagrosso (approved):
> Alright, I think I've addressed all the initial feedback and it's ready for another look (@colagrosso). I'll handle the merge conflict shortly.

Great job, Job! LGTM. I left a few nits before merging. Sorry I didn't see them before.

Removing the `GetArtist()` function from `Artwork.php` should be enough to handle the merge conflict.

> Regarding the autofill/search for existing artists, I think let's leave that for a separate PR after we've decided how it should be implemented. For this PR I'm happy to leave in the current datalist or take it out, whatever we think is appropriate.

Let's leave the datalist in add address it in a separate PR. You've done more than enough work here for one PR already. We'll add some details on #234 on what we learned here about artist help and any other issues. 

> Also, this PR has grown to contain quite a lot of commits. I'd normally squash it down to one or two before merging unless anyone would rather keep the entire history. Thoughts?

I appreciated all the small commits during the review because it was easy to see what changed, but now I think it would be better to squash it down to one or two before merging.

PR #238, opened 2023-06-17
Add Get() method to Artist

colagrosso:
Hi Job, I'm working on this admin page:

* `GET <root>/admin/artworks` -- view the queue of unapproved artwork

and I realized I made a mistake on the `Get()` function for `Artist.php`. I tested this PR locally and the `PropertiesBase` syntactic sugar does the right thing this way.

I peeked at your in progress PR #236, and this change doesn't look like it will conflict with yours. (Thanks for fixing some mistakes in `Artwork.php`, like adding `CompletedYearIsCirca`.)

Would you mind giving me a quick review before I merge?

cc: @jobcurtis 
Progress on #234

### DIFF
--- a/lib/Artist.php
+++ b/lib/Artist.php
@@ -21,6 +21,27 @@ class Artist extends PropertiesBase{
 			throw $error;
 		}
 	}
+	// ***********
+	// ORM METHODS
+	// ***********
+
+	public static function Get(?int $artistId): Artist{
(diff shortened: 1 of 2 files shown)


jobcurtis (approved):
Looks good to me :+1:

colagrosso:
Thanks, merging now.

PR #239, opened 2023-06-19
Don't store artwork tag URL in the database

colagrosso:
Hi, @jobcurtis, another short PR that shouldn't conflict with your work. It's a building block to the moderator admin page, which I'll also send out soon.

I realized I made a mistake and there's no good reason to store the artwork tag URL in the database. We should derive it from the name.

Progress on #234

### DIFF
--- a/config/sql/se/Tags.sql
+++ b/config/sql/se/Tags.sql
@@ -1,7 +1,6 @@
 CREATE TABLE `Tags` (
   `TagId` int(10) unsigned NOT NULL AUTO_INCREMENT,
-  `Name` varchar(80) NOT NULL,
-  `UrlName` varchar(80) NOT NULL,
+  `Name` varchar(255) NOT NULL,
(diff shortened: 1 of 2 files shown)


jobcurtis (approved):
One thing we could consider for storing slugged names is [generated columns](https://mariadb.com/kb/en/generated-columns/). It'd allow us to create indices against the slugged values which might be useful for the pages for individual artworks.

Probably an optimisation for later though - this looks good to me.

acabal:
I think in general we will want to avoid generated columns/triggers/SPs/other DB-level logic. DB should store data, while generating that data should be strictly app-level logic. Doing app logic in the DB may seem convenient at first but it inevitably becomes hard to reason about and manage in the long term.

colagrosso:
Thanks, both of you, merging now.

I also have some thoughts on how to implement this URL:

but we have a few more tasks to get to first.

PR #240, opened 2023-06-26
Artwork review queue for admins

colagrosso:
Hi @jobcurtis, here's a PR for you to try out and review. You'll need to use your form from #236 to submit some artwork, then visit:

```
/admin/artworks
```

to review them. After you've reviewed them, you can set your artwork records back to `unverified` in MySQL with:

```
MariaDB [se]> UPDATE Artworks SET Status = 'unverified' WHERE ArtworkId IS NOT NULL;
```

so that you can review them again.

This PR handles these two URLs:

* `GET <root>/admin/artworks` -- view the queue of unapproved artwork
* `POST <root>/admin/artworks/<artwork-id>` -- approve or reject artwork with status=<approved,declined>. Restfully this would be a PATCH request

One future TODO we'll have to note on #234 (in addition to any large issue you find): I did not yet password-protect the admin page via Apache.

### DIFF
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -130,9 +130,31 @@ class Artwork extends PropertiesBase{
 		}
 	}
 
-    /**
-     * @throws \Exceptions\ValidationException
-     */
+	// ***********
+	// ORM METHODS
+	// ***********
+
+	public static function Get(?int $artworkId): Artwork{
+		if($artworkId === null){
+			throw new Exceptions\InvalidArtworkException();
+		}
+
+		$result = Db::Query('
+				SELECT *
+				from Artworks
+				where ArtworkId = ?
+			', [$artworkId], 'Artwork');
+
+		if(sizeof($result) == 0){
+			throw new Exceptions\InvalidArtworkException();
+		}
+
+		return $result[0];
+	}
+
+	/**
+	 * @throws \Exceptions\ValidationException
+	 */
 	public function Create(): void{
 		$this->Validate();
 		$this->Created = new DateTime();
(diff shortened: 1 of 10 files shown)


lib/Artwork.php
```diff
@@ -168,6 +190,19 @@ public function Create(): void{
 		}
 	}
 
+	/**
+	 * @throws \Exceptions\ValidationException
+	 */
+	public function Save(): void{
```
jobcurtis, line 196: Suggestion: the name of this method could be more specific -- `SaveStatus` or something that indicates that it's only updating one field.

www/admin/artworks/post.php
```diff
@@ -0,0 +1,40 @@
+<?
+require_once('Core.php');
+
+if(HttpInput::RequestMethod() != HTTP_PATCH){
```
jobcurtis, line 4: More of a question than a request that this be changed: what is the purpose of the `_method` field and the check for `HTTP_PATCH` here?

I understand that it's a mechanism that already existed in the codebase, but it's not used anywhere else as far as I can tell. I suppose it indicates intent, but it's somewhat confusing at first glance since the actual request is still `POST` (and anyone seeing `PATCH` might assume that the request is being made by something other than an HTML form).

jobcurtis (commented):
Looks great! It's definitely all starting to come together. I've left a few comments; mostly just minor suggestions & queries. The only things that really need to be addressed are the invalid XHTML -- it seems that Firefox is more strict about it than Chromium.

lib/Artwork.php
colagrosso: I think the idea is that you could modify multiple properties on an object, then call `Save()` once (which would be just one database `UPDATE`).

[`NewsletterSubscription`](https://github.com/standardebooks/web/blob/5ea795881c86ca70b1292aba7185ef9a6928a555/lib/NewsletterSubscription.php#L73) can update three properties. [Here](https://github.com/standardebooks/web/blob/5ea795881c86ca70b1292aba7185ef9a6928a555/www/newsletter/subscriptions/post.php#L84) is an example of getting a NewsletterSubscription object, modifying it, and calling `Save()`.

colagrosso (commented):
Thanks for the speedy review!

lib/Artwork.php
jobcurtis: Fair enough - we can change the method if there are more properties that need to be updated.

jobcurtis (approved):
Looks good to me! Great work :+1:

PR #241, opened 2023-06-27
Browse approved artwork

colagrosso:
Implements these two URLs from #234:

* `GET <root>/artworks` -- browse the list of approved artwork

Depends on #240


colagrosso:
Job, I added a few enhancements to the `ArtworkDetail` template. They shouldn't affect the review too much if you've already started. Still no rush on this review if you haven't started yet.

www/admin/artworks/get.php
```diff
@@ -6,7 +6,7 @@
 try{
 	$artwork = Artwork::Get($artworkId);
 }
-catch(Exceptions\SeException $ex){
+catch(Exceptions\SeException){
```
jobcurtis, line 9:
```suggestion
catch(Exceptions\SeException $_){
```

I believe catching exceptions without assigning them to variables is only possible in PHP 8+. I'm not sure what `standardebooks.org` is actually running on but the README states 7+ and the included Dockerfile uses 7.4, so we should probably stick to syntax that is valid for PHP 7.

www/admin/artworks/post.php
```diff
@@ -11,7 +11,7 @@
 try{
 	$artwork = Artwork::Get($artworkId);
 }
-catch(Exceptions\SeException $ex){
+catch(Exceptions\SeException){
```
jobcurtis, line 14:
```suggestion
catch(Exceptions\SeException $_){
```

www/admin/artworks/post.php
```diff
@@ -31,10 +31,11 @@
 	}
 
 	$artwork->Save();
+	Library::RebuildArtworkCache();
 
 	http_response_code(303);
 	header('Location: /admin/artworks');
 }
-catch(Exceptions\SeException $ex){
+catch(Exceptions\SeException){
```
jobcurtis, line 39:
```suggestion
catch(Exceptions\SeException $_){
```

templates/ArtworkList.php
```diff
@@ -0,0 +1,20 @@
...
+	<li>
+		<div class="thumbnail-container">
+			<picture>
+				<a href="/artworks/<?= $artwork->Slug ?>"><img src="<?= $artwork->ThumbUrl ?>" property="schema:image"/></a>
+			</picture>
+		</div>
+		<p>Title: <a href="/artworks/<?= $artwork->Slug ?>" property="schema:name"><?= Formatter::ToPlainText($artwork->Name) ?></a></p>
+		<p>Artist: <span class="author" typeof="schema:Person" property="schema:name"><?= Formatter::ToPlainText($artwork->Artist->Name) ?></span></p>
+		<div>
+			<p>Year completed: <?= $artwork->CompletedYear ?></p>
```
jobcurtis, line 16: Suggestion: handle the 'unknown' and 'circa' cases here.

templates/ArtworkList.php
```diff
@@ -0,0 +1,20 @@
+<?
+
+$artworks = $artworks ?? [];
+?>
+<ol class="artwork-list list">
+<? foreach($artworks as $artwork){ ?>
+	<li>
+		<div class="thumbnail-container">
+			<picture>
+				<a href="/artworks/<?= $artwork->Slug ?>"><img src="<?= $artwork->ThumbUrl ?>" property="schema:image"/></a>
```
jobcurtis, line 10: nit: Strictly speaking a picture element "contains zero or more `<source>` elements and one `` element" (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture).

You could wrap the `a` around the entire `picture` element to make it valid HTML (even though all browsers will display it fine as is). The same thing is present in `ArtworkDetail.php` (which I didn't catch in the previous review, sorry).

jobcurtis (commented):
I've left a couple comments for minor issues. Nothing really to say about the overall functionality other than that it works very well!

www/admin/artworks/post.php
acabal: The server is now on PHP 8.1.2 so you can catch exceptions without assigning them to a variable.

lib/Library.php
```diff
@@ -159,6 +159,104 @@ public static function GetTags(): array{
...
+		switch($sort){
+			case SORT_COVER_ARTIST_ALPHA:
+				usort($matches, function($a, $b){
+					return strcmp(mb_strtolower($a->Artist->Name), mb_strtolower($b->Artist->Name));
+				});
+				break;
+
+			case SORT_COVER_ARTWORK_CREATED_NEWEST:
+				usort($matches, function($a, $b){
+					if($a->Created < $b->Created){
```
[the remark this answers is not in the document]
colagrosso: Ok, done.

(Background: I originally copied the patter in `FilterEbooks` in the file, e.g., [`lib/Library.php:76`](https://github.com/standardebooks/web/blob/ebecebe4e6bc43bd51d969bbbdb0c0a7c27f18c0/lib/Library.php#L76). I left that function intact, but flipped the comparison here.)

templates/ArtworkDetail.php
```diff
@@ -3,13 +3,36 @@
...
+		<td>Artist</td>
+		<td><?= Formatter::ToPlainText($artwork->Artist->Name) ?><? if($artwork->Artist->DeathYear !== null){ ?>, <abbr title="deceased">d.</abbr> <?= $artwork->Artist->DeathYear ?><? } ?></td>
+	</tr>
+	<tr>
+		<td>Completed Year</td>
+		<td><? if ($artwork->CompletedYear === null){ ?>(unknown)<? }else{ ?><?= $artwork->CompletedYear ?><? if($artwork->CompletedYearIsCirca){ ?> (circa)<? } ?><? } ?></td>
+	</tr>
+	<tr>
+		<td>File size</td>
+		<td><?= $artwork->ImageSize ?>
```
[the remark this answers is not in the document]
colagrosso: Thanks a bunch for catching that. Fixed.

templates/ArtworkList.php
colagrosso: Good call, done.

templates/ArtworkList.php
colagrosso: Yes, you're totally right. I made that change, which then also required some CSS changes. Those CSS changes sparked some other ideas to improve the `ArtworkList` style, so I think it's even better now.

colagrosso:
Ready for another look, Job. No rush.

www/admin/artworks/post.php
jobcurtis: Good to know - I'll see if I can get the docker image upgraded too.

jobcurtis (approved):
Looks good to me. I like the larger thumbnails in the list view!

PR #242, opened 2023-07-01
Use a foreach loop when inserting artwork tags

colagrosso:
Small change on #234


colagrosso:
@jobcurtis, fyi: I'm going to merge this without review because it's a small, nonfunctional change. It just corrects some poor style I used in a previous commit.

PR #243, opened 2023-07-02
Artwork submission form - persist fields when submission fails

jobcurtis:
Hey @colagrosso, here's a few small improvements to the artwork submission form, if you wouldn't mind having a look.

The first is that filled-in fields are persisted after failing validation. The second is less visible, but we're now cleaning up newly-created artists if validation fails for whatever reason. Together these changes required some changes to the order in which objects are created / captchas are checked, etc.

### DIFF
--- a/www/artworks/new.php
+++ b/www/artworks/new.php
@@ -109,12 +109,7 @@ if ($exception){
 						type="text"
 						name="artwork-tags"
 						placeholder="tags, comma-separated"
-						value="<?php
-							$tags = $artwork->ArtworkTags ?? [];
-							$tags = array_column($tags, 'Name');
-							$tags = implode(', ', $tags);
-							print $tags;
-						?>"
+						value="<?= $artwork->ArtworkTagsImploded ?>"
 					/>
 				</label>
 			</fieldset>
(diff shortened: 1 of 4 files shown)


www/artworks/new.php
```diff
@@ -64,20 +80,42 @@
...
 					Artwork Tags
-					<input type="text" name="artwork-tags" placeholder="tags, comma-separated"/>
+					<input
+						type="text"
+						name="artwork-tags"
+						placeholder="tags, comma-separated"
+						value="<?php
+							$tags = $artwork->ArtworkTags ?? [];
+							$tags = array_column($tags, 'Name');
+							$tags = implode(', ', $tags);
```
colagrosso, line 115: You could bundle up these three lines into a new method or property in `Artwork.php`. That way you could replace the logic here with something like:

```php
$artwork->ArrayTagsImploded
```

or

```php
$artwork->ArrayTagsString 
```

colagrosso (approved):
So much better, Job, thanks! Now I don't have to stress so much about making a mistake on the CAPTCHA. I added a few minor comments that you can overrrule or address in a subsequent PR. The discussion on #234 about what to pass into `Create()` might overlap or supersede these comments, so you can merge this PR first and then address that discussion and/or my comments.

PR #244, opened 2023-07-04
Align thumbnails in artwork list and match style in admin queue

colagrosso:


colagrosso:
Sorry for the noise above. This is a small change, so I'm going to submit it without review. I believe it's not conflicting with any ongoing work. 

@jobcurtis, FYI

PR #245, opened 2023-07-04
Add year of death to autocomplete to aid filling

colagrosso:
Hey, @jobcurtis, check this out. Recall our [previous discussion](https://github.com/standardebooks/web/pull/236#discussion_r1236509626) about making the artist autocomplete a little easier for the user. With this PR, the user can see the death year of the artists in the database in the autocomplete. It doesn't automatically fill in the "Year of death" field, but it makes it easier for the user to put the right year in there.

Here's what it looks like in action:
Screenshot 2023-07-04 12 24 16 AM

Progress on #234


jobcurtis:
It does look nice in Chrome. Unfortunately, Firefox displays just the inner contents of the `<option>` if it's specified, so this doesn't quite work there. (Excuse my garbage data)
firefox

colagrosso:
Well, we wouldn't want it to be too easy, would we? Seriously, though, thanks for checking.

What do you think of this version?
Screenshot 2023-07-04 1 54 01 PM

jobcurtis (approved):
I think that works. The name being there twice in Chrome is a little odd but I don't see a way around it.

PR #246, opened 2023-07-04
In use artwork: Style differently and link to ebook

colagrosso:
Check this out, @jobcurtis:

Screenshot 2023-07-04 1 36 59 AM

For artwork that is `in_use`, we style it differently, and it has a link to the ebook the artwork is used in.

If you'd like to modify your `Artworks` table in place and not drop and recreate it, you can run:
```sql
MariaDB [se]> ALTER TABLE Artworks ADD COLUMN EbookWwwFilesystemPath varchar(255) NULL AFTER ArtworkPage;
```

Progress on #234


colagrosso:
Small change to this PR: I made a common ArtworkStatus template because previously I had a complicated line copy–pasted in ArtworkDetail and ArtworkList.

jobcurtis (approved):
Looks good to me!

PR #249, opened 2023-07-08
Don't store UrlName in Artworks table

colagrosso:


colagrosso:
Hi @jobcurtis, I'm going to merge this PR without review. It's small and shouldn't conflict with what you're working on. Let me know if merges like this get in your way.

This PR removes the `UrlName` column from the `Artworks` table. I added it before I needed it, and it turns out we don't need it stored because

1. It's derived from `Name`
2. When we look something up by URL, we look in the APCu cache, not the database

If you have a database with a vestigial column, you can drop it with: 

```sql
MariaDB [se]> ALTER TABLE Artworks DROP COLUMN UrlName;
```

acabal:
UrlName should remain in the database. We need to check against it to locate the artwork to serve. APCu should not be used to index these artworks - that's what the database is for. The DB maintains its own query cache and the low expected traffic and small scale of the artworks site makes APCu irrelevant.

APCu is currently used for ebooks only because the system was designed before a DB made sense given the size of the project as a whole; in a sense, APCu *is* the DB for ebooks, as the ebook equivalent of UrlName is availalbe in the ebook metadata, and APCu stores the metadata DOM. If I were to start over today having to index 850 ebooks I would probably start with a DB and skip APCu entirely.

colagrosso:
Fair enough. Thanks for explaining. Tomorrow I'll revert this PR and change tactics.

PR #250, opened 2023-07-08
Wrap names and other text in ToPlainText()

colagrosso:
Hi @jobcurtis,

I had a little trouble filling out the submission form with special characters, failing the captcha, and then viewing the saved artwork and artist info ([relevant XKCD](https://xkcd.com/327/)):

Screenshot 2023-07-08 1 40 08 PM

So I wrapped the names and other text in `Formatter::ToPlainText()`. I might have gone overboard, or I might have missed some other cases, so check my work.

### DIFF
--- a/www/artworks/get.php
+++ b/www/artworks/get.php
@@ -11,7 +11,7 @@ if($artwork === null){
 	Template::Emit404();
 }
 
-?><?= Template::Header(['title' => Formatter::ToPlainText($artwork->Name)]) ?>
+?><?= Template::Header(['title' => $artwork->Name]) ?>
 <main class="artworks">
 	<section class="narrow">
 		<?= Template::ArtworkDetail(['artwork' => $artwork]) ?>
--- a/www/artworks/get.php
+++ b/www/artworks/get.php
@@ -11,7 +11,7 @@ if($artwork === null){
 	Template::Emit404();
 }
 
-?><?= Template::Header(['title' => $artwork->Name]) ?>
+?><?= Template::Header(['title' => Formatter::ToPlainText($artwork->Name)]) ?>
 <main class="artworks">
 	<section class="narrow">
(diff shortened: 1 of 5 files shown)


www/artworks/get.php
```diff
@@ -11,7 +11,7 @@
 	Template::Emit404();
 }
 
-?><?= Template::Header(['title' => $artwork->Name]) ?>
+?><?= Template::Header(['title' => Formatter::ToPlainText($artwork->Name)]) ?>
```
acabal, line 14: Notice that Template::Header *already* escapes the name. No need to escape it twice.

www/artworks/post.php
```diff
@@ -101,7 +101,7 @@ function parseArtworkTags(): array{
 
 	handleImageUpload($_FILES['color-upload']['tmp_name'], $artwork);
 
-	$_SESSION['success-message'] = '“' . $artwork->Name . '” submitted successfully!';
+	$_SESSION['success-message'] = '“' . Formatter::ToPlainText($artwork->Name) . '” submitted successfully!';
```
acabal, line 104: In general we only want to escape at the point where we output to HTML, which is typically only within a `<?= ?>` block. Escaping elsewhere invites accidental double-escapes, and also can lead to a problem where further processing before output expects unescaped text, but the text was surprisingly escaped earlier in processing, possibly even in a different function.

www/artworks/get.php
colagrosso: Noted, thanks. Reverted this change.

www/artworks/post.php
colagrosso: Good guideline about `<?= ?>` blocks, thanks. I changed the call to where this message is output to HTML.

jobcurtis (approved):
Good catch - thank you!

PR #251, opened 2023-07-08
Expand treatment of CompletedYear in AdminArtworkList

colagrosso:


colagrosso:
Another trivial change. Merging without review.

PR #252, opened 2023-07-09
Replace APCu with DB queries

colagrosso:
Hi @jobcurtis, this PR reverts #249. See the discussion about why we will store UrlName in the `Artworks` table. We'll need a UrlName in the `Artists` table, too, so this PR adds it. If you don't want to drop your `Artists` table, you can alter it:

```sql
ALTER TABLE Artists ADD COLUMN UrlName varchar(255) NULL AFTER Name;
```

@acabal: This change turned out to be easier than I thought. I think part of the feedback in #249 was to keep things straightforward and not add premature optimizations. That said, I did add an index on `Artworks.UrlName` in order to avoid table scans on `Artworks`. The query execution plan confirms it works correctly:

Without index on `Artworks.UrlName`:
```
MariaDB [se]> EXPLAIN SELECT Artworks.* from Artworks inner join Artists using (ArtistId) where Artists.UrlName = 'henri-fantin-latour' and Artworks.UrlName = 'night';                                         
+------+-------------+----------+--------+---------------+---------+---------+----------------------+------+-------------+
| id   | select_type | table    | type   | possible_keys | key     | key_len | ref                  | rows | Extra       |
+------+-------------+----------+--------+---------------+---------+---------+----------------------+------+-------------+
|    1 | SIMPLE      | Artworks | ALL    | NULL          | NULL    | NULL    | NULL                 | 63   | Using where |
|    1 | SIMPLE      | Artists  | eq_ref | PRIMARY       | PRIMARY | 4       | se.Artworks.ArtistId | 1    | Using where |
+------+-------------+----------+--------+---------------+---------+---------+----------------------+------+-------------+
```

With index on `Artworks.UrlName`:
```
+------+-------------+----------+--------+---------------+---------+---------+----------------------+------+-----------------------+
| id   | select_type | table    | type   | possible_keys | key     | key_len | ref                  | rows | Extra                 |
+------+-------------+----------+--------+---------------+---------+---------+----------------------+------+-----------------------+
|    1 | SIMPLE      | Artworks | ref    | index2        | index2  | 1023    | const                | 1    | Using index condition |
|    1 | SIMPLE      | Artists  | eq_ref | PRIMARY       | PRIMARY | 4       | se.Artworks.ArtistId | 1    | Using where           |
+------+-------------+----------+--------+---------------+---------+---------+----------------------+------+-----------------------+
```

As you wrote, with the low expected traffic and small scale of the artworks site, a table scan is probably no big deal, but the index seemed appropriate.

### DIFF
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -202,7 +202,7 @@ class Artwork extends PropertiesBase{
 		$this->Validate();
 		$this->Created = new DateTime();
 		Db::Query('
-			INSERT INTO Artworks (ArtistId, Name, CompletedYear, CompletedYearIsCirca, Created, MuseumPage,
+			INSERT INTO Artworks (ArtistId, Name, UrlName, CompletedYear, CompletedYearIsCirca, Created, MuseumPage,
 			                      PublicationYear, PublicationYearPage, CopyrightPage, ArtworkPage)
 			VALUES (?,
 			        ?,
@@ -213,8 +213,9 @@ class Artwork extends PropertiesBase{
 			        ?,
 			        ?,
 			        ?,
+			        ?,
 			        ?)
-		', [$this->Artist->ArtistId, $this->Name, $this->CompletedYear, $this->CompletedYearIsCirca,
+		', [$this->Artist->ArtistId, $this->Name, $this->UrlName, $this->CompletedYear, $this->CompletedYearIsCirca,
 				$this->Created, $this->MuseumPage, $this->PublicationYear, $this->PublicationYearPage,
 				$this->CopyrightPage, $this->ArtworkPage]
 		);
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -195,6 +195,21 @@ class Artwork extends PropertiesBase{
 		return $result[0];
 	}
 
+	public static function GetByUrlPath($artistUrlName, $artworkUrlName): Artwork{
+		$result = Db::Query('
+				SELECT Artworks.*
+				from Artworks
+				inner join Artists using (ArtistId)
+				where Artists.UrlName = ? and Artworks.UrlName = ?
+			', [$artistUrlName, $artworkUrlName], 'Artwork');
+
+		if(sizeof($result) == 0){
+			return null;
+		}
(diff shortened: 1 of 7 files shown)


lib/Artwork.php
```diff
@@ -195,14 +195,29 @@ public static function Get(?int $artworkId): Artwork{
 		return $result[0];
 	}
 
+	public static function GetByUrlPath($artistUrlName, $artworkUrlName): Artwork{
```
jobcurtis, line 198: I think we may want to think about how to ensure that this property is unique for each artwork. At the moment it's possible to have overlaps that make it impossible to access an artwork - or create links for one artwork that take you to a different one.

Here are some steps that reproduce this behaviour;

1. Submit artwork with artist name 'test-artist' & artwork name 'test-artwork'
2. Submit artwork with same artist/artwork name & a different image
3. In the artwork admin queue, decline the first entry and approve the second
4. In the artworks list, the second (approved) item will be present, but clicking on it will take you to the page for the first (declined) item.

It's a contrived example but I think this scenario could happen fairly easily if the first time an artwork is submitted it's rejected for missing proof of PD or incorrect metadata.

As for how to fix it, I don't have any perfect ideas at the moment. One way would be to check that there are no `approved` artworks with the same URL before allowing approval, and have this method filter by status as well.

n.b. I don't think this discussion needs to block this PR - I think the problem would have been present in the version using the cache as well.

jobcurtis (approved):

lib/Artwork.php
jobcurtis, line 198: nit: if it can return `null` then the return type of the function should be `?Artwork`

colagrosso: Thanks for noticing that. Let's address it in a follow-up PR. Your comment also reminded me to add a small commit (6d6258a) to this function that limits the query to Approved and In Use artwork. (Unverified and declined artwork can still be viewed on the admin page.) Despite that change, there's still a possibility of submitting two artworks with the same name and artist, which we will fix.

colagrosso:
Thanks for the review, Job, submitting now.

PR #253, opened 2023-07-15
Add artwork-specific CSS

colagrosso:
Hi, @jobcurtis, this refactor has a large diff, so check my work please.

I have another PR coming that includes some CSS changes, and I decided that we're polluting `core.css` with too much artwork-specific logic. The [SEMoS](https://standardebooks.org/manual) has its own CSS, so there's a precedent for this. Finally, I believe this will allow us to be more confident in refining the style of the artwork pages without worrying about breaking something on the rest of the site.

The crux of this PR moves artwork-specific CSS out of `core.css`. Here are the steps I followed to do that:
    
1. `git reset c335ad3 www/css/core.css www/css/dark.css`
2. Review the move the diffs from `core.css` and `dark.css` to `artwork.css` and `artwork-dark.css`

I reset `core.css` and `dark.css` back to c335ad3 (Feb 2023) because that is the last commit on those files before we started adding logic there. 

(Note that the master branch has a recent `core.css` change in fa952176d1e70d8fbdee908160b129e0346adc30. That's potentially another reason for us to not make `core.css` changes because it will make the merge easier later.)


www/css/artwork.css
```diff
@@ -0,0 +1,418 @@
...
+}
+
+main.artworks nav ol li.highlighted a,
+main.artworks nav ol li a:hover{
+	color: #fff;
+	text-decoration: none;
+	text-shadow: 1px 1px 0 rgba(0, 0, 0, .5);
+}
+
+input[type="number"],
```
jobcurtis, line 195: I'm not sure that it's very useful to move generic stuff like this out of `core.css`/`dark.css`. Now we have identical styles in two places, and they'll both need to be updated to stay consistent -- if just the `input[type="text"]` in `core.css` gets updated then we'll have differently-styled inputs in the artwork form. 

Also, if someone wants to use number or url inputs elsewhere in the site, they'd have to either move these changes back or recreate them in the main files.

colagrosso: Updated, thanks for weighing in. I deliberated on this, too, and decided you're right. It also leads fewer lines if CSS overall, which is good.

jobcurtis (approved):

PR #254, opened 2023-07-15
Add a Museums table

colagrosso:
Hi, @jobcurtis, this is the PR that I mentioned in #253 that includes some CSS changes.

The motivation from the PR came from a [previous discussion](https://github.com/standardebooks/web/issues/234#issuecomment-1587673991) about adding a `Museums` table and doing a simple check on the museum domain to make sure it's on the approved list.  Seemed useful, so I did it. 

I took the list of approved museums from [this list](https://standardebooks.org/manual/1.7.1/10-art-and-images#10.3.3.7.4) in SEMoS.


templates/ArtworkDetail.php
```diff
@@ -53,6 +53,18 @@
 <ul>
 	<li>Link to an approved museum page: <? if($artwork->MuseumPage !== null){ ?> <a href="<?= Formatter::ToPlainText($artwork->MuseumPage) ?>">Link</a><? }else{ ?>(not provided)<? } ?></li>
 </ul>
+<? if(!empty($artwork->MuseumPage)){ ?>
+<? $matchingMuseum = Museum::FindMatch($artwork->MuseumPage); ?>
+<? if($matchingMuseum !== NULL){ ?>
+<figure class="corrected full">
+	<p>Approved museum: <?= Formatter::ToPlainText($matchingMuseum->Name) ?> <code>(<?= Formatter::ToPlainText($matchingMuseum->Domain) ?>)</code>
```
jobcurtis, line 60: Missing a closing `</p>` on this line

lib/Museum.php
```diff
@@ -0,0 +1,26 @@
...
+			ORDER BY Name', [], 'Museum');
+	}
+
+	public static function FindMatch(string $url): ?Museum{
+		foreach(Museum::GetAll() as $museum){
+			if(stristr($url, $museum->Domain)){
+				return $museum;
+			}
+		}
+		return null;
```
jobcurtis, line 24: Minor suggestion: you could do the search in the database rather than fetching all possibilities and searching them in PHP. It's a very small set though so really comes down to personal preference.

```suggestion
		$result = Db::Query('
			SELECT *
			FROM Museums
			WHERE ? LIKE CONCAT("%", Domain, "%")
			LIMIT 1;
		', [$url], 'Museum');

		return $result[0] ?? null;
```

jobcurtis (approved):
Looks good, approved pending merge of #253 & that one missing `</p>`.

templates/ArtworkDetail.php
colagrosso: Great catch, thanks. Fixed.

lib/Museum.php
colagrosso: I like it. Updated. Thanks for proposing another way to do it.

My old `GetAll()` approach may have been better suited to using the APCu (or even just relying on the DB query cache), but this is less code.

colagrosso:
Job, this PR depends on the CSS in #253 , so I'll wait for your approval on that PR before submitting both.

PR #255, opened 2023-07-15
Move artwork creation logic from `artworks/post.php` to `Artwork.php`

jobcurtis:
Per [discussion](https://github.com/standardebooks/web/issues/234#issuecomment-1618593966), this PR moves construction of `Artwork` objects to the class itself.

The only functional changes are that tags/artists are no longer created if form validation fails. Also, failing to move/create image/thumbnail files will result in orphaned database objects (and a log message).

@acabal I've included you on this PR as it relates to your feedback.

### DIFF
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -151,6 +151,7 @@ class Artwork extends PropertiesBase{
 	// *******
 	// METHODS
 	// *******
+	/** @throws \Exceptions\ValidationException */
 	protected function Validate(): void{
 		$error = new Exceptions\ValidationException();
 
@@ -186,6 +187,19 @@ class Artwork extends PropertiesBase{
 
 		if($error->HasExceptions){
 			throw $error;
+		}
+	}
+
+	/** @throws \Exceptions\InvalidImageUploadException */
+	private function ValidateImageUpload(string $uploadPath): void{
+		$uploadInfo = getimagesize($uploadPath);
+
+		if ($uploadInfo === false){
+			throw new Exceptions\InvalidImageUploadException();
+		}
+
+		if ($uploadInfo[2] !== IMAGETYPE_JPEG){
+			throw new Exceptions\InvalidImageUploadException('Uploaded image must be a JPG file.');
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


colagrosso:
Right on, Job, thanks for following up on previous tasks! I tested it locally, and creating new artwork works as expected.

My feedback on the change: `Artwork::Build()` is effectively a constructor with a different name. Even though we [moved away](https://github.com/standardebooks/web/pull/235#issuecomment-1586495885) from using a big constructor previously, it seems appropriate now and moves logic out of `www/artworks/post.php` ("Fat Model, Skinny Controller"). LGTM, but wait for Alex.

jobcurtis:
Right, it is just a big constructor. I do think a constructor is appropriate here though, as it needs to interpret some arguments and the alternative would be creating class properties for use only during construction. Also, with PHP 8.0's named arguments, I'd say it's just as readable as assigning properties individually.

lib/Artwork.php
```diff
@@ -227,12 +241,102 @@ public static function GetByUrlPath($artistUrlName, $artworkUrlName): ?Artwork{
 		return $result[0];
 	}
 
+	public static function Build(string $artistName, ?int $artistDeathYear, string $artworkName, ?int $completedYear,
+				     bool $completedYearIsCirca, ?string $artworkTags, ?int $publicationYear,
+				     ?string $publicationYearPage, ?string $copyrightPage, ?string $artworkPage,
+				     ?string $museumPage): Artwork{
+		/** @return array<ArtworkTag> */
+		function parseArtworkTags(?string $artworkTags): array{
```
colagrosso, line 249: nit: I vote to move this to a top-level `protected static` function. I get that it's only called from `Build()`, but it makes the `Build()` function longer than necessary.

www/artworks/post.php
```diff
@@ -112,19 +42,7 @@ function parseArtworkTags(): array{
...
-	}
-
-	if (isset($artist->ArtistId)){
-		$artist->DeleteIfUnused();
-	}
-
+} finally{
 	http_response_code(303);
-	header('Location: /artworks/new');
+	header('Location: ' . '/artworks/new');
```
colagrosso, line 47: nit: Let's revert this line. The string concatenation isn't joining a string and a variable.

lib/Artwork.php
```diff
@@ -227,12 +241,102 @@ public static function GetByUrlPath($artistUrlName, $artworkUrlName): ?Artwork{
...
+		$this->Artist->GetOrCreate();
+		$this->Insert();
+
+		$storedImage = move_uploaded_file($uploadPath, WEB_ROOT . $this->ImageUrl);
+		$storedThumb = imagejpeg($thumbnail, WEB_ROOT . $this->ThumbUrl);
+
+		if (!($storedImage && $storedThumb)){
+			$log = new Log(ARTWORK_UPLOADS_LOG_FILE_PATH);
+			$log->Write("Failed to store image or thumbnail for uploaded artwork [$this->ArtworkId]");
+		}
```
colagrosso, line 308: Alex's [comment](https://github.com/standardebooks/web/issues/234#issuecomment-1618939776) in #234 about deleting incomplete state applies here. If one of these two file operations failed, we're in an incomplete state, so we should delete the orphan DB record created on line 300 above. The log statement is fine, but the code should also throw an exception to `post.php`.

colagrosso (commented):
Job, I previously wrote "LGTM, but wait for Alex," but I re-reviewed this PR and have two nits and an actual comment we should discuss. Please have a look, and apologies for not raising it sooner.

lib/Artwork.php
acabal: It's not necessary to create a DB record before we validate/manipulate images. Images can be validated first, and generated thumbs can go in temp files. Then after they're generated and validated, and everything else in the object passes validation, we create the DB record and then move all of the images into their final filenames which will be based on the DB primary key.

jobcurtis: OK, I've updated the `Create` method. It now does the following;

1. Validate
2. Create both images & store in temporary files (in the actual `images/uploads` directory)
3. Create database objects
4. Rename temporary files

The error handling on the final step consists of logging and warning the user that something went wrong. I think this is acceptable because a) there aren't really any realistic reasons why it would be possible to create the files but not rename them[^1], so this should never happen, and b) if it does happen, everything required to recover is available in the db, filesystem and logs.

[^1]: maybe if a file has already been created and had its permissions changed, which would have to have been done manually.

lib/Artwork.php
```diff
@@ -281,15 +272,43 @@ function parseArtworkTags(?string $artworkTags): array{
...
 		$this->Validate();
 		$this->ValidateImageUpload($uploadPath);
 
-		$thumbnail = self::GenerateThumbnail($uploadPath);
+		try{
+			$thumbPath = tempnam(WEB_ROOT . COVER_ART_UPLOAD_PATH, "tmp-thumb-");
+			$imagePath = tempnam(WEB_ROOT . COVER_ART_UPLOAD_PATH, "tmp-image-");
+
+			self::GenerateThumbnail($uploadPath, $thumbPath);
+			if (!move_uploaded_file($uploadPath, $imagePath)) throw new \Safe\Exceptions\FilesystemException;
```
colagrosso, line 305: nit: From the PHP code style at [README.md:177](https://github.com/standardebooks/web/blob/master/README.md?plain=1#L177), this should have braces:

```php
if (!move_uploaded_file($uploadPath, $imagePath)){
        throw new \Safe\Exceptions\FilesystemException;
}
```

colagrosso (approved):
LGTM, Job, thanks. I added one more nit, but otherwise I believe you've addressed all the feedback. The `Create()` function does all the work it can before inserting the DB record, and the rename step afterward is unlikely to fail, as you point out.

PR #257, opened 2023-07-19
Disallow duplicate artwork

colagrosso:
Job, following up on a [previous issue](https://github.com/standardebooks/web/issues/234#issuecomment-1636898313) you noted, this PR disallows these two situations:

1. Users cannot submit artwork that would match a URL of an existing approved or in-use artwork. That is, they can't submit something that would be a duplicate of something browsable. For example, they can't submit "Tramps" by Jules-Alexis Muenier if there is already artwork at `/jules-alexis-muenier/tramps`.
2. Admins can't approve unverified artwork in the same situation. (As you pointed out, this can happen by submitting two artworks with the same artist+name to the unverified state.)

Now a URL of the form `/artist-name/artwork-name` can match only one artwork. (As always, that artwork must be approved or in use.)

There are two small, related refactors in this PR, too:

1. Updated this `Save()` function to accept a `status` parameter to match what you did with `Build()` and `Create()`
2. Replaced the `Slug` property with a more useful `Url` property

### DIFF
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -4,7 +4,7 @@ use function Safe\filesize;
 
 /**
  * @property string $UrlName
- * @property string $Slug
+ * @property string $Url
  * @property array<ArtworkTag> $ArtworkTags
  * @property string $ArtworkTagsImploded
  * @property Artist $Artist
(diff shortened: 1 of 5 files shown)


jobcurtis (approved):
Looks good - approved.

With this I think we actually have the beginnings of an (awkward and very manual) fixup workflow for admins - you can go to the admin page for an existing artwork (you have to work this out from the image url), decline it, create and approve a new artwork with updated details. If you didn't change the artwork or artist name, the original URL now points to your updated version.

Obviously it should be streamlined in the future with an actual edit feature, but this is useful to know about in the meantime.

colagrosso:
Right on, Job, thanks for the review. Merging now.

Good thinking on the admin workflow. I can imaging a combination of operations like that and requesting someone with enough privilege to run something manually. As you point out, still should be streamlined in the future.

PR #258, opened 2023-07-22
Add alternate spellings for Artists

colagrosso:
Alex mentioned this in a [Nov 2021 discussion](https://groups.google.com/g/standardebooks/c/Y-0WRq1Gmvo/m/vWXY54cKBgAJ):

> Fuzzy search is necessary because often artists had their names romanized in different ways over the decades (Ilya Repin, Ílya Repín, Illia Riepin, Илья Репин?) 

With this PR the user can search via alternate spellings. For now, the alternate spellings must be inserted via SQL, e.g., 

```
MariaDB [se]> INSERT INTO AlternateSpellings VALUES (46, 'Ílya Repín'), (46, 'Illia Riepin'), (46, 'Илья Репин');
```

but once they are inserted the search on alternate spellings works as expected:

Screenshot 2023-07-22 1 32 25 PM

### DIFF
--- a/config/sql/se/AlternateSpellings.sql
+++ b/config/sql/se/AlternateSpellings.sql
@@ -1,0 +1,5 @@
+CREATE TABLE `AlternateSpellings` (
+  `ArtistId` int(10) unsigned NOT NULL,
+  `AlternateSpelling` varchar(255) NOT NULL,
+  UNIQUE KEY `idxUnique` (`ArtistId`,`AlternateSpelling`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/lib/Artist.php
+++ b/lib/Artist.php
@@ -3,12 +3,14 @@ use Safe\DateTime;
 
 /**
  * @property string $UrlName
+ * @property array<string> $AlternateSpellings
  */
 class Artist extends PropertiesBase{
 	public $ArtistId;
 	public $Name;
 	public $DeathYear;
 	protected $_UrlName;
+	protected $_AlternateSpellings;
 
 	// *******
 	// GETTERS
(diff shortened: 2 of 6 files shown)


jobcurtis (approved):
Sure, this looks good :+1:

This isn't strictly related to this PR, but I've been thinking about adding fuzzy search functionality that doesn't require maintaining a list of alternate spellings using some combination of `soundex()` and MySQL's full text search. The basic idea would be that each database object gets a column storing a text index in soundex representation - e.g. Ílya Repín becomes `I400 R150` (if we use the sanitised url name). We could then generate a query string by concatenating the same representation to the user's query (so a search for 'Illia Riepin' becomes `Illia Riepin I400 R150`). A query like

```sql
SELECT * 
FROM Artists 
WHERE MATCH (Name, SoundexName) AGAINST ( 'Illia Riepin I400 R150' );
````

will return the correct result, accounting for soundalike alternate spellings and minor typos.

We'd still need the alternate spellings for different alphabets, or where there are large enough differences. We'd also need to move some of the search functionality into SQL queries rather than keeping it in PHP code as it is now. I think we'd get quite a bit of extra functionality without having to add much data though.

What do you think? Would this be worth me spending a bit of time on to produce a prototype?

jobcurtis:
The soundex algorithm was actually developed for proper names (for use in census analysis according to Wikipedia). It works best with English pronunciation, but a lot of alternative spellings are anglicised versions of the original name, so it does work pretty well on this kind of data.

It might not be perfect but I think it's worth a try. If it doesn't work well enough we can remove it or try something else. PHP has a built-in `metaphone()` which works similarly but is supposed to be more accurate for English words. We could also try calculating similarity (PHP has `levenshtein()` and `similar_text()` or we could roll our own) but you can't generate indices for similarity so this way would be a lot more efficient if it worked.

acabal:
OK, you can give it a shot if you want.

colagrosso:
Thanks for the review and discussion, everyone. Merging now.

PR #260, opened 2023-08-06
Update wanted-ebooks.php

SzymonSzott:
The Mystery of the Yellow Room by Gaston Leroux has already been released.


acabal:
If it has already been released, you can just delete it from the page completely. Thanks!

PR #261, opened 2023-08-06
Change if statement to match code style

colagrosso:
Previously mentioned here: https://github.com/standardebooks/web/pull/255#discussion_r1268786970


colagrosso:
Trivial change, merging without review.

PR #262, opened 2023-08-06
Use ArtworkList for admin and non-admin views

colagrosso:
Job, way back in #241 you [pointed out](https://github.com/standardebooks/web/pull/241#discussion_r1248860048) a fix for `Template::ArtworkList` that I applied, but forgot to copy the same fix to `Template::AdminArtworkList`. I now think that we should have just one template, which is what this PR does.

As we've added more features like tags and alternate artist spellings, it's not worth maintaining two templates that do the same thing.

It's a bit hacky how I pass an extra parameter into the template, but I think it's ok. Let me know if you can think of a way that will read better, even if it's just better formatting or variable names.


jobcurtis (approved):
Looks good, I think the alternative to passing in `useAdminUrl` would be to pass an anonymous function that gets the right url given an Artwork. But there's not much difference in the approaches; this is fine too.

colagrosso:
Thanks for the review and the discussion. Merging now.

PR #264, opened 2023-08-13
Fulltext search with phonetic index on artist names

jobcurtis:
Hey @colagrosso, I was talking about some ideas for a more forgiving search functionality on an earlier PR; here's an implementation of one of those ideas. It uses the `metaphone` function to generate a phonetic representation for the artist name, which in theory would cover some typos and minor alternative spellings. It also uses mysql/mariadb fulltext indices for query matching.

This is mostly experimental so I would be interested to hear your thoughts -- definitely not set on merging this exactly as it is now. My first impressions are that it returns a lot more results, so it's good if you don't know exactly what you're searching for. On the other hand, if you have a specific query it will return a lot of less specific results - for example, searching 'portrait woman' will return all images that match either 'portrait' or 'woman'.

You can also try different phonetic keys which give somewhat different results (just replace references to `metaphone` with `soundex`). In general I found that soundex gives more false positives (it matches 'childe' and 'claude'), but metaphone has more false negatives ('ilia' and 'illia' match, but not 'ilya').

Migration scripts:

```sql
ALTER TABLE Artists
	ADD PhoneticName varchar(255) NOT NULL;

ALTER TABLE Artists
	ADD FULLTEXT ArtistsFullText (Name, UrlName, PhoneticName);

ALTER TABLE AlternateSpellings
	ADD FULLTEXT AlternateSpellingsFullText (AlternateSpelling);

ALTER TABLE Artworks
	ADD FULLTEXT ArtworksFullText (Name, UrlName);

ALTER TABLE Tags
	ADD FULLTEXT TagsFullText (Name);
```

```php
{
	/** @var Artist $artist */
	foreach (Artist::GetAll() as $artist) {
		$metaphone = explode(separator: '-', string: $artist->UrlName);
		$metaphone = array_map(callback: 'metaphone', array: $metaphone);
		$metaphone = implode(separator: ' ', array: $metaphone);
		Db::Query('
			UPDATE Artists
			SET PhoneticName = ?
			WHERE ArtistId = ?;
		', [$metaphone, $artist->ArtistId]);
	}
}
```


colagrosso:
Great work, Job! Way to prototype your idea. You know database features that I don't.

Here are my thoughts:

1. I vote to delay changes to search until we get, say, 500 approved images. Once the list of artwork spans multiple pages, we can collect feedback from users, i.e., book producers. We could ask broadly on the mailing list or tap specific people whose opinion we trust. I'd like to hear from book producers how important it is to cover some typos and minor alternative spellings. While it sounds useful, the need may be small and be obviated by sites like Wikipedia. Moreover, the consequence of returning too many non-matching results might mean the search is too hard to understand what it's doing.

(Side note: I bet we'll get more feedback about too many images having too few tags. I expect book producers to browse by tag just as often searching by artwork title and artist name.)

2. When it's time to revisit search, I think we should consider your two PR's two contributions separately:

a. Replace the current approach of loading all artwork and filtering it ([`Artwork::GetBrowsable()`](https://github.com/standardebooks/web/blob/d9edcd568c2f9ee07194d616b822ed8fd6e380ff/lib/Artwork.php#L431C25-L431C37), [`Artwork::Contains()`](https://github.com/standardebooks/web/blob/d9edcd568c2f9ee07194d616b822ed8fd6e380ff/lib/Artwork.php#L438), and [`Library::FilterArtwork()`](https://github.com/standardebooks/web/blob/d9edcd568c2f9ee07194d616b822ed8fd6e380ff/lib/Library.php#L168C25-L168C38)) with a direct SQL query

Here are the pros/cons I see of replacing the current approach with a direct SQL query:

+ Pro: Probably the "right" way to do it
+ Pro: Slightly fewer lines of code
+ Pro: Might be easier to test in MySQL directly
- Con: The implementation in this PR has a gnarly query with a large number of joins. (The current approach also has a large number of lookups. They're just behind `PropertiesBase` methods.). Performance is not a serious concern, but we should at least run EXPLAIN on the query to ensure it behaves as expected. I'd be more concerned about maintaining the query if we needed to add more search parameters or sort orders.

One possible pro of the current approach:
+ The code works like ebooks search, which might be easier to maintain

b. Adding phonetic matches

I'm skeptical that we'll need this, but I'd be convinced if our users ask for it.

----

One immediate improvement that your PR uncovered: We should move `Library::FilterArtwork()` out of `lib/Library.php` into `lib/Artwork.php`. Now that we're not using APCu for artwork (317bac2), it doesn't need to be near the other APCu functions. 

This probably isn't the response you were expecting or looking for, but I do appreciate you and your willingness to prototype. Great initiative!

acabal:
My suggestion is not to overengineer this project at this stage. The important thing is getting an MVP out so we can make sure the basics work and that it meets user expectations. Then, we can go back and chew on things like improved queries, indexes, making sure phonetics work, etc.

jobcurtis:
Thanks both for your feedback. Closing this PR now. I agree that this type of stuff isn't entirely necessary at this stage -- maybe something to come back to in the future.

PR #267, opened 2023-08-19
Remove "Title: " and "Author: " from ArtworkList

colagrosso:
They are still in the ArtworkDetail table, but this cleans up the list a bit. The meaning should be clear from the context, especially with the improved style that matches ebook treatment.

Here is a before and after:

Screenshot 2023-08-19 4 09 04 PM

Screenshot 2023-08-19 4 05 02 PM


jobcurtis (approved):

colagrosso:
Thanks for the review, Job. Merging now.

PR #268, opened 2023-08-20
Update text to reflect changes in Dr Jekyll versions on PG

vr8hub:
A lot has changed with PG's Dr Jekylls since the guide was originally written: the one that used to be the most downloaded is now outpaced 8-1 by the one the guide uses, and the biggest problem stated by the guide (page numbers) are no longer present. In order to not have to change the wording if the situation changed again, I just referred to the versions generically (rather than "most/least popular"), left the reason for the choice (cleaner transcription), but updated why we thought it was cleaner.

See what you think; there isn't a lot of difference between the two versions now other than what's stated here; the version we're using has less commas, and more modern compound words (preoccupied rather than pre-occupied, anyone vs any one, etc.).


acabal:
Yes, I think that works. Thanks!

PR #272, opened 2023-09-10
Updated wanted ebooks list: Norwegian Folktales

bentley:
Asbjørnsen and Moe seem to be the Brothers Grimm of Norway, both in action and in reputation.

Gutenberg’s translations are by Dasent, and were described favorably by the Norwegian authors ([Wikipedia](https://en.wikipedia.org/wiki/Norwegian_Folktales#Translation_into_English)).


acabal:
Thanks, I'll add this now. Normally we reserve additions to the wanted list as a Patrons Circle membership perk. Otherwise anyone can add things and it becomes a big list of random books instead of something more focused.

PR #273, opened 2023-09-11
Added new How To page showing how to handle color lines in maps

drgrigg:
This was sparked by this discussion in the Forum here: https://groups.google.com/g/standardebooks/c/OOntbKGyAT8/m/y4inBzoBCAAJ. Hope it's useful!


acabal:
Thanks David. I think we should update the instructions to use a free cross platform SVG editor like Inkscape. Not everyone uses Macs (I don't) and not everyone should need to buy specific software to follow a how-to. Can you do this in Inkscape?

drgrigg:
I guess you'll also want me to use GIMP, then, instead of Acorn? That's going to be a big ask!

acabal:
GIMP is for raster images, so if necessary yes. Inkscape is for SVGs. Inkscape can trace bitmaps, I don't think you're going to need GIMP.

drgrigg:
It's the colour replacement step which I'm looking to a bitmap editor for. Acorn is brilliant for this, but I've been struggling for the last hour to see how to do it in GIMP. It's possible, but it doesn't do anywhere near as good a job as Acorn. Still, I'll persevere.

drgrigg:
OK, I've changed everything to use GIMP and Inkscape. 

As I've added in a note at the bottom, Inkscape does have a “multi-color” trace function, which in theory ought to eliminate the steps we needed to carry out in GIMP but in practice I haven’t been able to get this to work in any satisfactory manner. If someone is more familiar than me with Inkscape and can get it to work well, then I could eliminate almost all of the steps I've described.

acabal:
Excellent David, thanks!

PR #274, opened 2023-09-24
Use `disabled="disabled"` in `/admin/artworks/get.php`

jobcurtis:
This is required by Firefox when rendering XHTML pages


colagrosso:
Good fix, Job. Thanks for being diligent about testing on Firefox. This is a quick fix, so I'm going to merge it without further delay. Sorry for the slow review.

PR #275, opened 2023-09-24
Set `upload_max_filesize` to 128M to allow larger image uploads.

jobcurtis:
The default is 2M, which is much too small for many images from museum websites.

Also added feedback when a user attempts to upload an image that is too large.

### DIFF
--- a/config/php/fpm/standardebooks.org.ini
+++ b/config/php/fpm/standardebooks.org.ini
@@ -10,6 +10,9 @@ allow_url_fopen = false
 allow_url_include = false
 expose_php = Off
 
+post_max_size = 128M
+upload_max_filesize = 128M
+
 [Date]
 date.timezone = Etc/UTC
 
--- a/config/php/fpm/standardebooks.org.ini
+++ b/config/php/fpm/standardebooks.org.ini
@@ -10,8 +10,8 @@ allow_url_fopen = false
 allow_url_include = false
 expose_php = Off
 
-post_max_size = 128M
-upload_max_filesize = 128M
+post_max_size = 64M
+upload_max_filesize = 32M
 
 [Date]
 date.timezone = Etc/UTC
--- a/www/artworks/post.php
+++ b/www/artworks/post.php
@@ -31,6 +31,16 @@ try{
 		throw new Exceptions\InvalidCaptchaException();
 	}
 
+	if ($_FILES['color-upload']['error'] > 0){
+		// see https://www.php.net/manual/en/features.file-upload.errors.php
+		$message = match ($_FILES['color-upload']['error']){
+			1 => 'Image upload too large (maximum ' . ini_get('upload_max_filesize') . ')',
+			default => 'Image failed to upload',
+		};
+
+		throw new \Exceptions\InvalidImageUploadException($message);
+	}
+
 	$artwork->Create($_FILES['color-upload']['tmp_name']);
 
 	$_SESSION['success-message'] = '“' . $artwork->Name . '” submitted successfully!';
--- a/www/artworks/post.php
+++ b/www/artworks/post.php
@@ -9,7 +9,28 @@ if (HttpInput::RequestMethod() != HTTP_POST){
 
 session_start();
 
+function post_max_size_bytes(): int{
+	$post_max_size = ini_get('post_max_size');
+	$unit = substr($post_max_size, -1);
+	$size = (int) substr($post_max_size, 0, -1);
+
+	return match ($unit) {
+		'g', 'G' => $size * 1024 * 1024 * 1024,
+		'm', 'M' => $size * 1024 * 1024,
+		'k', 'K' => $size * 1024,
+		default => $size
+	};
+}
+
 try{
+	if (empty($_POST) || empty($_FILES)){
+		if ($_SERVER['CONTENT_LENGTH'] > post_max_size_bytes()){
+			throw new \Exceptions\InvalidRequestException("Request too large (maximum " . ini_get('post_max_size') . ')');
+		} else{
+			throw new \Exceptions\InvalidRequestException();
+		}
+	}
+
 	$artwork = Artwork::Build(
 		artistName: HttpInput::Str(POST, 'artist-name', false),
 		artistDeathYear: HttpInput::Int(POST, 'artist-year-of-death'),
@@ -31,11 +52,12 @@ try{
 		throw new Exceptions\InvalidCaptchaException();
 	}
 
-	if ($_FILES['color-upload']['error'] > 0){
+	$uploadError = $_FILES['color-upload']['error'];
+	if ($uploadError > 0){
 		// see https://www.php.net/manual/en/features.file-upload.errors.php
-		$message = match ($_FILES['color-upload']['error']){
+		$message = match ($uploadError){
 			1 => 'Image upload too large (maximum ' . ini_get('upload_max_filesize') . ')',
-			default => 'Image failed to upload',
+			default => 'Image failed to upload (error code ' . $uploadError . ')',
 		};
 
 		throw new \Exceptions\InvalidImageUploadException($message);
--- a/www/artworks/post.php
+++ b/www/artworks/post.php
@@ -2,13 +2,6 @@
 
 require_once('Core.php');
 
-if (HttpInput::RequestMethod() != HTTP_POST){
-	http_response_code(405);
-	exit();
-}
-
-session_start();
-
 function post_max_size_bytes(): int{
 	$post_max_size = ini_get('post_max_size');
 	$unit = substr($post_max_size, -1);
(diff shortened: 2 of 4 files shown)


colagrosso:
Hi, Job, sorry for the delay. I'll keep reviewing this, but I'll show you what I'm seeing so far, even though I don't know how to fix it yet. 

1. With the changes to `config/php/fpm/standardebooks.*.ini`, I can now upload large images. Yay!
2. When I remove the changes, or set them to something lower like

```
post_max_size = 10M
upload_max_filesize = 10M
```

I don't get an error message in the browser. Instead, the submit form silently fails and returns to the blank form. In the error log (`/var/log/local/www-error.log`), I see:

```
[Sun Oct 01 03:41:09.358363 2023] [proxy_fcgi:error] [pid 2060823:tid 139873101006400] [client 
73.243.245.57:46552] AH01071: Got error 'PHP message: PHP Warning:  POST Content-Length of 14399805 
bytes exceeds the limit of 10485760 bytes in Unknown on line 0PHP message: PHP Fatal error:  Uncaught 
TypeError: Artwork::Build(): Argument #1 ($artistName) must be of type string, null given, called in 
/standardebooks.org/web/www/artworks/post.php on line 24 and defined in 
/standardebooks.org/web/lib/Artwork.php:280\nStack trace:\n#0 
/standardebooks.org/web/www/artworks/post.php(24): Artwork::Build()\n#1 {main}\n  thrown in 
/standardebooks.org/web/lib/Artwork.php on line 280', referer: https://143.198.245.16/artworks/new
```

It looks like something is failing before your filesize checking is kicking in. I'll keep trying to make sure I'm not missing something.

jobcurtis:
Hmm, good catch. I guess I was mostly testing with `upload_max_filesize` and added `post_max_size` later. It appears that `post_max_size` gets checked before `upload_max_filesize`, and PHP doesn't really provide a way of handling requests that are too large.

I've added feedback for when the request is too large, but I'm not very happy with it. For one thing, we can't restore the form values (because PHP doesn't even store them). A better user experience would be to set a very high `post_max_size` (or unsetting it with `post_max_size = 0`), so that we hit the validation around uploaded files instead. I'm not really sure of the implications of allowing very large requests for the entire site though, so I'll probably leave that up to @acabal.

acabal:
Unfortunately PHP has very weak support for checking for requests that are too large. All we can do is set some reasonable anti-spam upper bound like 64MB, and make the publicly advertised limit much lower, like 32MB. 

That way the PHP script will receive any file under 64MB, and we can check the file size using PHP code to ensure its below our advertised limit of 32MB. However files over 64MB will be killed by PHP-FPM before the script is even executed, and we can't do anything about that, that's just how PHP works.

acabal:
Also note that post_max_size and upload_max_filesize are subtly different: https://stackoverflow.com/questions/23686505/php-post-max-size-vs-upload-max-filesize-what-is-the-difference

colagrosso:
+1 to setting the public limit to 32MB. If this data helps: Out of the several hundred of `cover.source.jpg` files across all ebook git repos, only 9 are bigger than 32MB.

```
$ ls -lSh /standardebooks.org/web/www/images/uploads
total 3.0G
-rw-rw-r-- 1 se       se         99M Sep  4 16:49 389.jpg
-rw-rw-r-- 1 se       se         99M Sep  4 16:48 214.jpg
-rw-rw-r-- 1 se       se         61M Sep  4 16:50 504.jpg
-rw-rw-r-- 1 se       se         51M Sep  4 16:50 579.jpg
-rw-rw-r-- 1 se       se         49M Sep  4 16:52 788.jpg
-rw-rw-r-- 1 se       se         41M Sep  4 16:48 224.jpg
-rw-rw-r-- 1 se       se         41M Sep  4 16:47 164.jpg
-rw-rw-r-- 1 se       se         36M Sep  4 16:47 84.jpg
-rw-rw-r-- 1 se       se         33M Sep  4 16:52 739.jpg

-rw-rw-r-- 1 se       se         28M Sep  4 16:49 446.jpg
-rw-rw-r-- 1 se       se         28M Sep  4 16:49 331.jpg
-rw-rw-r-- 1 se       se         26M Sep  4 16:53 877.jpg
-rw-rw-r-- 1 se       se         23M Sep  4 16:53 858.jpg
-rw-rw-r-- 1 se       se         20M Sep  4 16:50 564.jpg
-rw-rw-r-- 1 se       se         19M Sep  4 16:52 794.jpg
-rw-rw-r-- 1 se       se         19M Sep  7 15:56 987.jpg
-rw-rw-r-- 1 se       se         19M Sep  4 16:52 848.jpg
-rw-rw-r-- 1 se       se         18M Sep  4 16:51 673.jpg
-rw-rw-r-- 1 se       se         17M Sep  4 16:51 676.jpg
-rw-rw-r-- 1 se       se         16M Sep  4 16:46 71.jpg
-rw-rw-r-- 1 se       se         16M Sep  4 16:49 418.jpg
-rw-rw-r-- 1 se       se         15M Sep  4 16:50 554.jpg
-rw-rw-r-- 1 se       se         15M Sep  4 16:51 594.jpg
-rw-rw-r-- 1 se       se         15M Sep  4 16:52 778.jpg
-rw-rw-r-- 1 se       se         15M Sep  4 16:52 816.jpg
-rw-rw-r-- 1 se       se         15M Sep  4 16:47 183.jpg
-rw-rw-r-- 1 se       se         14M Sep  4 16:53 889.jpg
[...]
```

www/artworks/post.php
```diff
@@ -9,7 +9,28 @@
...
+	return match ($unit) {
+		'g', 'G' => $size * 1024 * 1024 * 1024,
+		'm', 'M' => $size * 1024 * 1024,
+		'k', 'K' => $size * 1024,
+		default => $size
+	};
+}
+
 try{
+	if (empty($_POST) || empty($_FILES)){
```
colagrosso, line 26: nit: no space after if, here and below

www/artworks/post.php
```diff
@@ -9,7 +9,28 @@
...
+		if ($_SERVER['CONTENT_LENGTH'] > post_max_size_bytes()){
+			throw new \Exceptions\InvalidRequestException("Request too large (maximum " . ini_get('post_max_size') . ')');
+		} else{
```
colagrosso, line 29: nit: no space before else, just

```
}else{
```

www/artworks/post.php
colagrosso: (There are a few other cases of this across the code that I'll clean up in a followup PR.)

www/artworks/post.php
```diff
@@ -31,6 +52,17 @@
 		throw new Exceptions\InvalidCaptchaException();
 	}
 
+	$uploadError = $_FILES['color-upload']['error'];
+	if ($uploadError > 0){
+		// see https://www.php.net/manual/en/features.file-upload.errors.php
+		$message = match ($uploadError){
+			1 => 'Image upload too large (maximum ' . ini_get('upload_max_filesize') . ')',
```
colagrosso, line 59: Replace the `1` with `UPLOAD_ERR_INI_SIZE` and remove the comment above with the link to php.net.

```
UPLOAD_ERR_INI_SIZE => 'Image upload too large (maximum ' . ini_get('upload_max_filesize') . ')',
```

config/php/fpm/standardebooks.org.ini
```diff
@@ -10,6 +10,9 @@ allow_url_fopen = false
 allow_url_include = false
 expose_php = Off
 
+post_max_size = 128M
+upload_max_filesize = 128M
```
colagrosso, line 14: Given the conversation so far, I vote for these values (here and below):

```
post_max_size = 64M
upload_max_filesize = 32M
```

That will restrict the total post data to 64M and each individual file to 32M. 

1. If the user uploads a file above 64M, they will lose their form values. Too bad. They will at least get an error, thanks to your new code.
2. If the user uploads a file between 32M and 64M, they will get an error, and the form will preserve their values, so that's nice.

www/artworks/post.php
```diff
@@ -9,7 +9,28 @@
...
+		'm', 'M' => $size * 1024 * 1024,
+		'k', 'K' => $size * 1024,
+		default => $size
+	};
+}
+
 try{
+	if (empty($_POST) || empty($_FILES)){
+		if ($_SERVER['CONTENT_LENGTH'] > post_max_size_bytes()){
+			throw new \Exceptions\InvalidRequestException("Request too large (maximum " . ini_get('post_max_size') . ')');
```
colagrosso, line 28: When we report the error to the user we should tell them the advertised maximum file size, i.e.:

```
ini_get('upload_max_filesize')
```

colagrosso (commented):
I was able to try this out with different values of `post_max_size` and `upload_max_filesize` to trigger your error messages. Good job! A few comments, then I think I'd like to try it one more time before approving.

www/artworks/post.php
colagrosso: (I suppose you can keep the comment above that links to php.net if you'd like. The point was to use the `UPLOAD_ERR_INI_SIZE` constant directly and not depend on the link to know what the `1` meant.)

www/artworks/post.php
```diff
@@ -9,7 +9,28 @@
 
 session_start();
 
+function post_max_size_bytes(): int{
```
colagrosso, line 12: sorry, another nit: Consider moving this function to the top of the file under the `require_once` line.

colagrosso (commented):
Another comment

www/artworks/post.php
colagrosso, line 28: nit: single quotes to be consistent

```
'Request too large (maximum `
```

jobcurtis:
Thanks for your review; I've updated the PR with the changes you suggested.

colagrosso:
LGTM

I tested it, and it works well. Thanks, Job!

colagrosso (approved):

PR #277, opened 2023-10-03
Update wanted ebooks list

erinendrei:
Hume's _Enquiry Concerning Human Understanding_ has been approved twice as a first production and abandoned. It seems like it'd be an appealing one for a new producer


acabal:
Thanks!

PR #279, opened 2023-10-05
Step-by-step guide: Add guidance to use /artworks

colagrosso:
Lots of eyeballs on this guide, so I'm open to different wording. I thought it was important to guide producers to `/artworks` early in the artwork step.


acabal:
We don't want to do this yet, the site isn't even live or beta ready yet. We should launch it first, test it with some experienced producers, make fixes and updates, then we can suggest it to the general public.

colagrosso:
Fair enough. Getting more feedback sounds great. Holding back this change a bit longer is fine.

PR #280, opened 2023-10-05
Review form: Trim trailing whitespace on status

colagrosso:
This is a bit of a hack, but there are several places in `Template::ArtworkStatus` that can introduce trailing whitespace, and I'd rather leave that template formatted as-is, so I added a `trim()` call where it's needed when using the status in a sentence. (Using the status in the table above is fine.)

### DIFF
--- a/templates/ArtworkStatus.php
+++ b/templates/ArtworkStatus.php
@@ -3,14 +3,8 @@
 $artwork = $artwork ?? null;
 ?>
 <? if($artwork !== null){ ?>
-	<? if($artwork->Status === "approved"){ ?>
-		Approved
-	<? }else if($artwork->Status === "in_use"){ ?>
-		In use
-		<? if($artwork->Ebook !== null && $artwork->Ebook->Url !== null){ ?>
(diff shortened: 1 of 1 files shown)


colagrosso:
This isn't a great example, but here's a example of the problem: The text reads, "Artwork cannot be approved because J.J. Shannon, R.A. by Orlando Rouland exists with status: Approved ." with a trailing space.

Screenshot 2023-10-05 1 06 48 AM

acabal:
This should be fixed within the template, not outside of the template. We should not need to tweak the output of template code before outputting.

colagrosso:
Ok, take 2. Fixed within the template, and still readable.

jobcurtis (approved):
I would almost say that this could just be a plain old method `Artwork::StatusFormatted` but this works too.

colagrosso:
Thanks for the review, Job, merging now.

> I would almost say that this could just be a plain old method `Artwork::StatusFormatted` but this works too.

Good point. I'll keep that in mind if we update the artwork status further.

PR #281, opened 2023-10-07
Fix artist name sort for names with diacritics

colagrosso:
Without the fix, names like Édouard Manet and Élisabeth Louise Vigée Le Brun were sorted at the end of the list when choosing the sort order "Artist name (a → z)".

Here's a screenshot of it working with the fix:

Screenshot 2023-10-07 12 20 39 AM

### DIFF
--- a/lib/Library.php
+++ b/lib/Library.php
@@ -203,9 +203,17 @@ class Library{
 
 		switch($sort){
 			case SORT_COVER_ARTIST_ALPHA:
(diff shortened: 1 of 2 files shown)


acabal:
Thanks!

acabal:
Oops, I see this was for cover art and not for the general website. Sorry!

colagrosso:
No prob! I used the same sorting approach as other examples in `lib/Library.php`.

PR #282, opened 2023-10-09
Fix errors found by PHPStan

colagrosso:
Alex made a similar fix in bd4b8d8944ef45069f6b983288a372ca79423ff4 on Jun 21, and I reused his approach from `lib/Ebook.php` for this line in `lib/Artwork.php`:

```php
$factor = intval(floor((strlen((string)$bytes) - 1) / 3));
```

One other benefit of running PHPStan: I found an actual, user-facing bug. Before these fixes, I was able to submit artwork with just spaces as the artist or artwork name. The submission would fail silently, and the error log contained:

```
Got error 'PHP message: PHP Fatal error:  Uncaught TypeError: Artwork::Build(): Argument #1 ($artistName) must be of type string, null given, called in /standardebooks.org/web/www/artworks/post.php on line 48 and defined in /standardebooks.org/web/lib/Artwork.php:274\nStack trace:\n#0 /standardebooks.org/web/www/artworks/post.php(48): Artwork::Build()\n#1 {main}\n  thrown in /standardebooks.org/web/lib/Artwork.php on line 274'
```

Now we check for valid artist and artwork name before calling `Artwork::Build()` and return an error to the user if there's a problem.

I also included a separate commit for some code style fixes that accumulated over the last few months. e.g., `if (` -> `if(`. I'm responsible for lots of them, so no judgment.

### DIFF
--- a/config/phpstan/phpstan.neon
+++ b/config/phpstan/phpstan.neon
@@ -28,3 +28,6 @@ parameters:
 		- DONATION_ALERT_ON
 		- DONATION_DRIVE_ON
 		- DONATION_DRIVE_COUNTER_ON
+	earlyTerminatingMethodCalls:
+		Template:
+			- Emit404
--- a/lib/Artist.php
+++ b/lib/Artist.php
@@ -112,7 +112,7 @@ class Artist extends PropertiesBase{
 			WHERE Name = ? AND DeathYear = ?
 		', [$this->Name, $this->DeathYear], 'Artist');
 
-		if (isset($result[0])){
+		if(isset($result[0])){
 			$this->ArtistId = $result[0]->ArtistId;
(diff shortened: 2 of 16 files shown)


jobcurtis (approved):
Good catch, approved. Something for the future might be to introduce 'safe' versions of the `HttpInput` methods (maybe `HttpInput::RequireStr`).

colagrosso:
Thanks for the review, Job, merging now.

> Good catch, approved. Something for the future might be to introduce 'safe' versions of the `HttpInput` methods (maybe `HttpInput::RequireStr`).

Yeah, with a fair bit of refactoring, that could lead to overall shorter code in files like `www/artworks/post.php`.

PR #283, opened 2023-10-09
Artwork review: Explain existing/approved artwork

colagrosso:
There are two similar but different cases to explain to reviewers:

1. The reviewer is looking at the admin page of artwork that has already been approved or is in use. They can't approve it again.

2. The reviewer is looking at unverified artwork that has the same title and artist as a different approved or in use artwork. They can't approve it because it would create a duplicate.

Here are two screenshot examples of the cases:

1. 
Screenshot 2023-10-08 10 18 26 PM

2.
Screenshot 2023-10-08 10 23 18 PM

I'm open to rewording these.

### DIFF
--- a/www/admin/artworks/get.php
+++ b/www/admin/artworks/get.php
@@ -19,8 +19,10 @@ catch(Exceptions\SeException){
 	<h2>Review</h2>
 	<? if($existingArtwork === null){ ?>
 	<p>Review the metadata and PD proof for this artwork submission. Approve to make it available for future producers.</p>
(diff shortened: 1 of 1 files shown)


jobcurtis (approved):
I would suggest removing the Status parenthetical from 

```php
already approved or in use (Status: <?= Template::ArtworkStatus(['artwork' => $existingArtwork]) ?>).
```

because it is already in the artwork detail section, or making it just say either "is already approved" or "is already in use". Not a requirement though - it's definitely fine as is.

colagrosso:
Great suggestion, Job. It's cleaner. I took it and am merging now.

PR #287, opened 2023-10-16
Admin approved message: Add link to artwork

colagrosso:
This is useful because after approving an artwork, there's not an easy way to get to the public, shareable link of that artwork.

Here's what the new approved message looks like:

Screenshot 2023-10-15 9 46 23 PM

where the link points to `/artworks/john-singer-sargent/oyster-gathering`.

The declined message is unchanged:

Screenshot 2023-10-15 10 03 28 PM

For the declined message, I considered adding a link to `/admin/artworks/<artwork-id>`, but decided not to.

### DIFF
--- a/www/admin/artworks/index.php
+++ b/www/admin/artworks/index.php
@@ -5,12 +5,20 @@ use function Safe\session_unset;
 
 session_start();
 
-$approvedMessage = $_SESSION['approved-message'] ?? null;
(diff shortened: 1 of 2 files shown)


jobcurtis (approved):
Looks good!

colagrosso:
Thanks for the review. Merging now.

PR #288, opened 2023-10-16
Insert Artists with DeathYear populated

colagrosso:
It bothers me that more than 90% of the artists have their death year unknown. This can be seen when submitting new artwork or browsing existing artwork. It sends the wrong message to users that we don't care about the accuracy of the data.

So I looked up all the ones that I could, which turned out to be 555 of them. That gives us a good foundation with a lot of artists that future submitters will be able to reuse. 

There are still a few unknown death years where I gave up my search, and even though I tried to be careful it's possible I made mistakes. Also, the database will be a moving target and require regular maintenance as submitters add missing or inconsistent death years. Still, I think this data is worth adding. 

Here is the before and after:

Screenshot 2023-10-11 10 20 15 PM

Screenshot 2023-10-15 11 36 00 PM

Here are the unknown death years I gave up on for now:

```
> SELECT * FROM Artists WHERE DeathYear IS NULL ORDER BY Name;
+----------+---------------------------+---------------------------+-----------+
| ArtistId | Name                      | UrlName                   | DeathYear |
+----------+---------------------------+---------------------------+-----------+
|      502 | Cecilia C. Foster         | cecilia-c-foster          |      NULL |
|      520 | Cornelius Gijsbrechts     | cornelius-gijsbrechts     |      NULL |
|      238 | Frank Wiles               | frank-wiles               |      NULL |
|      351 | Gilbert Sackerman         | gilbert-sackerman         |      NULL |
|      298 | Jan Augustin van der Goes | jan-augustin-van-der-goes |      NULL |
|      329 | Jean-Baptiste Messier     | jean-baptiste-messier     |      NULL |
|      295 | Joseph Kallaus            | joseph-kallaus            |      NULL |
+----------+---------------------------+---------------------------+-----------+
```
Here is a fancier table that shows which painting we have for those artists and what ebook it's used in:

```
> SELECT a.Name, a.DeathYear, substring(art.Name, 1, 30) AS `Artwork Name`, art.CompletedYear, substring(art.EbookWwwFilesystemPath, 1, 80) AS EbookWwwFilesystemPath FROM Artists a LEFT JOIN Artworks art USING (ArtistId) WHERE a.DeathYear IS NULL ORDER BY a.Name;
+---------------------------+-----------+-----------------------------------+---------------+----------------------------------------------------------------------------------+
| Name                      | DeathYear | Artwork Name                      | CompletedYear | EbookWwwFilesystemPath                                                           |
[...]

PR #289, opened 2023-10-16
Insert alternate spellings

colagrosso:


colagrosso:
Ok, @jobcurtis, I decided to take the "actual" spelling from the artists' Wikipedia titles. I also made this file consistent with the artists in #288, which required adding a few more rows. This is ready for review.

jobcurtis (approved):
Looks good to me. If we're serious about keeping duplicates down in the future then one of the things we might want is a tool for merging duplicate artists. I think it would be pretty straightforward to implement. Probably worth waiting to see how many duplicates we actually get though.

colagrosso:
Thanks for the review, merging now.

> If we're serious about keeping duplicates down in the future then one of the things we might want is a tool for merging duplicate artists. I think it would be pretty straightforward to implement. Probably worth waiting to see how many duplicates we actually get though.

Yep, agreed. I've done it manually a few times:

1. Identify the canonical artist with `CanonicalArtistId1` and alternate spelling artist with `AlternateArtistId2`. Those are the only two inputs needed.
2. `UPDATE Artworks SET ArtistId = CanonicalArtistId1 WHERE ArtistId = AlternateArtistId2;`
3. `INSERT INTO AlternateSpellings VALUES (CanonicalArtistId1, (SELECT Name FROM Artists WHERE ArtistId = AlternateArtistId2));`
4. `DELETE FROM Artists WHERE ArtistId = AlternateArtistId2;`

A tool or a short script could run the steps a bit quicker.

PR #290, opened 2023-10-22
Fix typo 'musuem' -> 'museum'

jobcurtis:
@colagrosso FYI - I think this PR is too small to worry about reviewing, so I'll merge it straight away.


colagrosso:
Great catch, and good call merging it right away!

PR #291, opened 2023-10-22
Rebase covers onto master and merge

colagrosso:
This brings all the changes from the `covers` branch into the `master` branch and adds two commits at the end to make `covers` consistent with `master`:

1. Covers files: SeException -> AppException
2. Covers files: Remove unneeded Core.php lines

After merging these changes, continue with Step 4 of this [2023-10-10 post](https://github.com/standardebooks/web/issues/234#issuecomment-1754622926) on issue #234, i.e., "Add the new MySQL tables."


acabal:
Great! I'm extremely busy this week so I'll have to wait till next week to take a look. If everything goes smoothly we can launch it for our editors to play with.

colagrosso:
Right on, Alex! Thanks for the detailed review and sorry about the confusion that will need some refactoring to fix. Your outline above makes sense. I'll get on this.

colagrosso:
I addressed the minor comments first, then I fixed the bug you found and took a whack at the refactor.

If you need to repeat any steps in the [Oct 10 instructions](https://github.com/standardebooks/web/issues/234#issuecomment-1754622926), I edited them:

* Step 4: Get the `insert` data from [this gist](https://gist.github.com/colagrosso/a2931d4bb730913d0a334448cb6ce7cf)
* Step 5: New ZIP file of unverified artwork (same artwork, but the code in the ZIP file needed a refactor to remove `Build()`)

The fix for the bug you found: The code now won't create a DB record if it detects that permissions problem with the `www-data` user. It was actually pretty subtle:

1. `tempnam()` was trying to create a file at `WEB_ROOT . COVER_ART_UPLOAD_PATH`. When it doesn't have permission there, PHP falls back to creating the file in `/tmp` and logs a notice: `PHP Notice:  tempnam(): file created in the system's temporary directory`. I didn't know about that behavior. 
2. The `GenerateThumbnail()` and `move_uploaded_file()` steps "succeed" by operating on files in `/tmp`.
3. __DB record gets created__
4. The `rename()` step to `WEB_ROOT . $this->ImageUrl` failed because of permissions problems. This step is after the DB record is created.

I fixed the bug by adding a validation error if `tempnam()` can't create a file in the right place.

Speaking of validation errors, I was able to get multiple child validation errors to appear if the submission has multiple problems:

Screenshot 2023-12-10 12 46 16 AM

Tried to upload a PNG here:
Screenshot 2023-12-10 12 51 01 AM

Permissions problems trigger these errors:
Screenshot 2023-12-10 1 55 19 AM

acabal:
OK excellent, thanks. It would actually be best to keep temp files in the system's temp folder, because if something goes wrong we won't litter our local dir with zombie temp files. At least in system temp we're sure that the system will clean zombies up at some point. So we should explicitly use system temp and only `rename()` once we're sure we're done with everything.

colagrosso:
Thanks for the quick reply!

Just to recap/restate the problem: I'd like to keep the design where the thumbnail and image filenames are based on the artwork ID, but the consequence is that there will always be a file operation after `Insert()`, which is a little risky.

This has been a handy design because it means that we don't need another field in the `Artworks` table to store the path to the thumbnail/image. It's also easy to see what's going on in the `WEB_ROOT . COVER_ART_UPLOAD_PATH` directory (i.e., `/standardebooks.org/web/www/images/uploads`). Because I'd like to keep this design, here's what I propose:

### Current code

The current code does this:

1. `Validate()`

a. Creates the thumbnail in `COVER_ART_UPLOAD_PATH` with a temporary name
b. Moves the uploaded image to `COVER_ART_UPLOAD_PATH` with a temporary name

2. `Insert()`

Inserts the DB record, which also gets the artwork ID via:

`$this->ArtworkId = Db::GetLastInsertedId();`

3. Last step in `Create()`

Renames the temporary files which are already in `COVER_ART_UPLOAD_PATH`, e.g., if `ArtworkId` equals 1210, rename them to `1210.jpg` and `1210.thumb.jpg`. It's possible this step fails, but unlikely.

### Proposal

We could change the code to use the system's temp folder this way:

1. `Validate()`

a. Generate the thumbnail in the system's temp folder
b. Check that `WEB_ROOT . COVER_ART_UPLOAD_PATH` is writable with `is_writable()`, but don't write anything there yet

2. `Insert()`

Same as above.

3. Last step in `Create()`

Move the thumbnail and image to `COVER_ART_UPLOAD_PATH` with the correct file name. It's possible this step fails, but unlikely.

I think this proposal is actually what [you wrote](https://github.com/standardebooks/web/pull/255#discussion_r1268509580) back on Jul 19. I'll start to make this change, so holler if I'm on the wrong track.

acabal:
Yes, that's basically what I wrote earlier. If you're saying there's a separate `Artwork::Insert()` method, that's not necessary, just do an insert query in `Artwork::Create()`.

colagrosso:
Ok, done. I made the change above to use the system's temp folder, and I also removed the `Artwork::Insert()` method that  wasn't necessary and put the insert query in `Artwork::Create()`.

I gave it a quick test of:

* the submission form
* the script to add 56 unverified artworks
* the script to add `in_use` artwork from published ebooks 

and it seemed to work fine.

I also double-checked that the bug you found is still fixed: If the `www-data` user doesn't have permission to write to `/standardebooks.org/web/www/images/uploads`, the submission form presents the validation error to the user.

acabal:
Looks great! A few more notes:

- I'm able to submit artwork with no tags. But, we want to require tags, browsing by tags is one of the big benefits of this database.
- In `Artwork::Validate()` you add the same one exception to the validation output in all cases. But look at my [example above](https://github.com/standardebooks/web/pull/291#issuecomment-1828511298) - instead of adding one exception, there is a new exception for every error. For example, if there are no tags, we get a `TagsRequired()` exception, not an `InvalidArtworkException('Tags required')`. This allows us to reuse generic exceptions... we can reuse this exception again if for example we later want to tag artists. This does mean creating a lot of tiny objects but files are free, and generic ones like TagsRequired can be reused later.
- When creating the artwork, you validate it before validating and creating the artist. This is a problem because what if the artist name is blank? Simply use your browser's web inspector to remove the `required` attribute from the artist name and submit the form - it crashes. Again, see my example above - the artist is validated first, *as part of the artwork's validation function*, because an artwork with an invalid artist, is itself invalid!

Can you sort all that out and do another round of testing?

We cannot assume that what we get via POST will be prevalidated by the form itself. Remove all the `required` attributes for testing and play around to make sure your validation function is robust against any kind of input.

colagrosso:
Right on, thanks for the extra feedback! Sorry I missed some of the points you made in your code example above. 

I'll get right on these changes and do another round of testing, but I just wanted to briefly mention something about this point:

> I'm able to submit artwork with no tags. But, we want to require tags, browsing by tags is one of the big benefits of this database.

Agreed. I will make that change. Even with the 56 examples I started with, I like browsing by tags.

Additional note: I'm also loving that I can search and browse the 800+ artworks from existing productions. Those artworks don't have tags, so I'll add validation logic in such a way that `unverified` artwork requires tags but `in_use` artwork doesn't. (`unverified` means "submitted via the form," and `in_use` means "from existing productions, added via script as an optional step in `deploy-ebook-to-www`.") This is similar to how artwork from existing productions doesn't have PD proof tracked in the DB, but submitted artwork of course requires PD proof.

acabal:
I also want to point out that having complete input validation isn't an academic matter. People could be using browsers that don't support the `required` attribute (lots of Windows XP machines still out there especially globally, or people on basic non-Chrome mobile browsers), or they could have a browser addon that messes with forms. This is a public form, so script kiddies can and will poke at it to see how it breaks including submitting raw POST requests that bypass our HTML form entirely. Therefore we have to validate everything carefully server side and output appropriate error messages for each case.

colagrosso:
Yep, understood. Thanks for sharing what I'm sure were hard won lessons.

I was lax on validation during development, but we'll clean it up.

colagrosso:
Ok, just for testing I removed all the all the `required` attributes, which made it easier to add full server-side validation. Validation is now performed both in the web browser and in the `Artist::Validate()` and `Artwork::Validate()` methods. There are three broad groups of text input fields in the submission form:

1. Three required strings: Artist Name, Artwork Name, Artwork Tags

Each one is required, but no other restrictions on format other than they must be short enough to fit in the DB.

2. Three years: Artist DeathYear, Artwork CompletedYear, PD proof PublicationYear

None of these are required, but if they are present they must be greater than zero and less than or equal to the current year. If the user bypasses the validation in the web browser and inputs a non-integer string, `HttpInput::Int()` converts the input to the default value, which is null. It would be nicer to raise a validation exception, but it's not possible to handle missing and invalid inputs differently in `HttpInput::Int()` as it's written today. Because these years are optional, we have to allow missing inputs to be set to the default null value.

3. Four URLs: MuseumUrl, PublicationYearPageUrl, CopyrightPageUrl, ArtworkPageUrl

None of these are required either, but if they are present they must be a valid URL. I'm using:

```php
filter_var($url, FILTER_VALIDATE_URL)
```

to check the URL format. It works fine, but I'm open to other suggestions.

I tried various empty, whitespace, numeric, etc. inputs where they didn't belong, and the validation caught them.

acabal:
Great thanks. Some bugs/notes:

- If I enter an artist name who already exists, but a different death year, a new row is created with the same `UrlName`. Therefore they are indistinguishable. If we enter an artist whose `UrlName` already exists, then we should use the existing row and ignore the different death date.
- If I add two artworks for one artist, but both artworks have the same `UrlName`, this succeeds. Again, the artworks will be indistinguishable so we should throw an error if the artwork name already exists for that particular artist.
- The important point here is that `UrlName` is how we must distinguish rows, not `Name`, because the `UrlName` is what appears in the URL and must be globally unique. This also affects `Artist::GetOrCreate()`.
- If I check the 'circa' box and submit the form with errors, the box is unchecked when the errors are printed
- The file upload input should accept other kinds of images like PNG, TIFF, BMP as museums often offer downloads in those formats. So we should be able to accept them and also process them. The thumbnails we generate for use on the website can be strictly JPG, but the original file should be stored in its original format. We will probably have to add a `MimeType` column to the `Artworks` table to support this. Also note that we must validate that the file we receive is actually a supported format, the filename extension is not enough to tell us as, since anyone can change it. (Renaming `virus.exe` to `virus.png` and uploading it for example).
- In `Artwork.php` line 226 you check for the max length of the imploded tag string. But, this check tells us nothing, because what we're *actually* interested is the length of each individual tag, since each tag is a row, not the length of all of the tags combined into one string. So you should drop the `GetGetArtworkTagsImploded` function entirely and check the length of each tag to make sure each tag is not too long for the `Tags` table.
- In `Artwork::Create()`, what is the `copyFile` parameter for? It looks like it's only used by a one-time script, `upsert-to-cover-art-database`. Instead of enshrining a parameter for a one-off script, just copy by default in the function, and delete the file in the script itself (which is basically a move)
- `Artwork::GetBrowsable` should be put somewhere else, probably in `Library.php`. An Artwork should not return an array of Artworks. But a Library could return an array of Artworks.
- `Artwork::Save()` should save the entire object, not just a few fields. If we want to use `Save()` in the future for other purposes it will be very surprising that it only saves a few things.

acabal:
If adding non-JPG support is too much work right now then let's shelve that for the moment, we can add that later. Let's hammer out the actual bugs first.

colagrosso:
> * If I enter an artist name who already exists, but a different death year, a new row is created with the same `UrlName`. Therefore the two rows are now indistinguishable. If we enter an artist whose `UrlName` already exists, then we should use the existing row and ignore the different death date.

Done. Much cleaner.

Side note: I'm not clear what the purpose is of the `Artists.DeathYear` field. I thought it could be used to distinguish between two artists that have the same name, but I guess that is much less likely to occur than a user neglecting to put the correct death year in the submission form.

> * If I add two artworks for one artist, but both artworks have the same `UrlName`, this succeeds. Again, the two artworks will be indistinguishable so we should throw an error if the artwork `UrlName` already exists for that particular artist.

This one is different from the one above: `UrlName` isn't reserved for an artwork until its `Status` is `approved` or `in_use`. Those are the only artworks that are available via URL. `unverified` and `declined` are not.

For example, if the status of the existing artwork is `declined` because it was poor quality or incorrect PD proof, we still allow a future submission. Once an artwork is `approved` or `in_use`, then no other submissions with that `UrlName` are allowed. We check for that in `Artwork::Validate()` (granted I did just fix a small bug there):

```php
         // Check for existing Artwork objects with the same URL but different Artwork IDs.                                                                                                                                                                                      
         $existingArtwork = Artwork::GetByUrlPath($this->Artist->UrlName, $this->UrlName);                                                                                                                                                                                       
         if($existingArtwork !== null && ($existingArtwork->ArtworkId !== $this->ArtworkId)){                                                                                                                                                                                    
                $error->Add(new Exceptions\ArtworkAlreadyExistsException(SITE_URL . $existingArtwork->Url));                                                                                                                                                                    
         }                            
```

which leads to this error message on submission:

Screenshot 2023-12-16 1 18 52 AM

If you quickly create two submissions with the same `UrlName`, they both go to the `unverified` state, but the moderator can approve only one of them. That gives them the opportunity to check the quality of each submission. On the second one they'll see:

Screenshot 2023-12-16 1 10 11 AM

I hope duplicate submissions aren't common, because I don't want to waste the moderator's time. Unless removing duplicates becomes a major source of tedious work, I believe we're already handling artwork `UrlName` correctly.


> * The important point here is that `UrlName` is how we must distinguish rows, not `Name`, because the `UrlName` is what appears in the URL and must be globally unique. This also affects `Artist::GetOrCreate()`.

For artists: I updated the `Artists` table to make `UrlName` a unique key and changed `Artist::GetOrCreate()` to use it.

For artworks: We have an index on `UrlName` and use it for looking up an artwork. It's not a unique key like it is for artists because of what I wrote above about artwork status, but I believe it's true to your intention: There is only one publicly available (`approved` or `in_use`) artwork per `UrlName` allowed.

> * If I check the 'circa' box and submit the form with errors, the box is unchecked when the errors are printed

Yikes, sorry about that. Fixed.


> * In `Artwork.php` line 226 you check for the max length of the imploded tag string. But, this check tells us nothing, because what we're _actually_ interested in is the length of each individual tag, since each tag is a row, not the length of all of the tags combined into one string. So you should drop the `GetArtworkTagsImploded` function entirely and check the length of each tag to make sure each tag is not too long for the `Tags` table.

Fair enough. That was a bit of a meaningless shortcut. I figured the number of tags would be small, each one would be short, and that the validation was there just to guard against abuse. We now check that each tag fits in the `Tags` table, and I added a limit of 100 tags per artwork. I'm up for changing the number, but I think there should be a limit.

We still need the `GetArtworkTagsImploded` function for its original purpose: If the user gets a validation error on submission, we put the imploded string back in the submission form.

> * In `Artwork::Create()`, what is the `copyFile` parameter for? It looks like it's only used by a one-time script, `upsert-to-cover-art-database`. Instead of enshrining a parameter for a one-off script, just copy by default in the function, and delete the file in the script itself (which is basically a move)

Sorry, that was an ugly wart. Removed. 

There's still some tension in `Artwork::Create()` because `move_uploaded_file()` seems to be required for files submitted via HTTP POST. If I try to use `copy()` on files submitted via HTTP POST, I get a permission denied error:

```
AH01071: Got error 'PHP message: PHP Warning:  copy(/standardebooks.org/web/www/images/uploads/544.jpg): Failed to open stream: Permission denied in /standardebooks.org/we
b/vendor/thecodingmachine/safe/generated/filesystem.php on line 114', referer: https://10.0.1.4/standardebooks.test/artworks/new
```

but I might be missing something. It doesn't work the other way, either. `move_uploaded_file()` won't work on files that weren't uploaded via HTTP POST.

For now, I'm using `is_uploaded_file()` to distinguish when I can use `move_uploaded_file()` for uploads and `copy()` for the script. That got rid of the need for the `copyFile` parameter. 

Side note: `upsert-to-cover-art-database` is not a one-time script. I'd like `upsert-to-cover-art-database` to be called by default by `deploy-ebook-to-www`. Right now, it's a flag with the default set to off. Once we're comfortable that it works correctly, it will keep up with all the `in_use` artwork from published ebooks. If a publisher used one of the `approved` artworks from the site, the script will update its status to `in_use`.

> * `Artwork::GetBrowsable` should be put somewhere else, probably in `Library.php`. An Artwork should not return an array of Artworks. But a Library could return an array of Artworks.

Good call, done.

> * `Artwork::Save()` should save the entire object, not just a few fields. If we want to use `Save()` in the future for other purposes it will be very surprising that it only saves a few things.

Got it. Done.

> * The file upload input should accept other kinds of images like PNG, TIFF, BMP as museums often offer downloads in those formats. So we should be able to accept them and also process them. The thumbnails we generate for use on the website can be strictly JPG, but the original file should be stored in its original format. We will probably have to add a `MimeType` column to the `Artworks` table to support this. Also note that we must validate that the file we receive is actually a supported format, the filename extension is not enough to tell us as, since anyone can change it. (Renaming `virus.exe` to `virus.png` and uploading it for example).

I haven't started on this one yet, but I'll work on it next unless something else comes up in your next round of feedback.

jobcurtis:
The original implementation of the image handling did handle other formats before we decided to limit it to JPG only. If you haven't started on that yet @colagrosso I am happy to resurrect it & add TIFF/BMP support as I have some time this weekend.

On a related note; what would be the purpose of storing mimetype in the database? As far as I understand it's not needed for serving the files, and if we do need it for anything in the future we can retrieve it from the file extension (which we are setting ourselves).

P.S. Thanks @colagrosso for all your work handling feedback so far.

colagrosso:
Sure, Job, go for it. Thanks!

Probably obvious, but work directly from this branch, please, because
`covers` is way behind.

acabal:
> On a related note; what would be the purpose of storing mimetype in the database? As far as I understand it's not needed for serving the files, and if we do need it for anything in the future we can retrieve it from the file extension (which we are setting ourselves).

You'd store the mime type because otherwise when rendering a page you wouldn't know in advance what the filename would be, because the filename itself is not stored in the DB.  To render a page with the correct download link you would need to generate the expected filename with ArtworkId + mime type extension.

acabal:
> This one is different from the one above: UrlName isn't reserved for an artwork until its Status is approved or in_use. Those are the only artworks that are available via URL. unverified and declined are not.

OK. I think the logic should be to deny a work with the same UrlName as an initial validation error, not to be left to the moderator. This hints to the user that the same artworks is already in the queue and doesn't need to be submitted again. This is also less work for the moderator who doesn't need to review duplicate entries; and forcing the UrlName to be unique is more obvious from a code maintenance perspective.

> There's still some tension in Artwork::Create() because move_uploaded_file() seems to be required for files submitted via HTTP POST. If I try to use copy() on files submitted via HTTP POST, I get a permission denied error:

That's fine, use `move_uploaded_file()` then. We don't really care what happens to the temp files once we're done with them, they're either gone (moved) or in system temp so not our problem any more.

If you both think we can get BMP/TIFF/PNG support in soonish, then I'll wait for that to happen. Otherwise let's wrap up these last bits and this is probably ready for initial release.

colagrosso:
> > This one is different from the one above: UrlName isn't reserved for an artwork until its Status is approved or in_use. Those are the only artworks that are available via URL. unverified and declined are not.
> 
> OK. I think the logic should be to deny a work with the same UrlName as an initial validation error, not to be left to the moderator. This hints to the user that the same artworks is already in the queue and doesn't need to be submitted again. This is also less work for the moderator who doesn't need to review duplicate entries; and forcing the UrlName to be unique is more obvious from a code maintenance perspective.

Fair enough, done.

> > There's still some tension in Artwork::Create() because move_uploaded_file() seems to be required for files submitted via HTTP POST. If I try to use copy() on files submitted via HTTP POST, I get a permission denied error:
> 
> That's fine, use `move_uploaded_file()` then. We don't really care what happens to the temp files once we're done with them, they're either gone (moved) or in system temp so not our problem any more.

Thanks. I left the `Artwork::Create()` code as-is, even if it's a little verbose. It correctly handles uploaded files via HTTP POST with `move_uploaded_file()` and files on the filesystem via the script with `copy()`, and there's no extra parameter. 

`Artwork::Create()` is probably worth another look from you and @jobcurtis. It does a lot.

> If you both think we can get BMP/TIFF/PNG support in soonish, then I'll wait for that to happen. Otherwise let's wrap up these last bits and this is probably ready for initial release.

Thoughts, Job?

I vote to package up what we have into an initial release.

jobcurtis:
> You'd store the mime type because otherwise when rendering a page you wouldn't know in advance what the filename would be, because the filename itself is not stored in the DB.

Ah yes. I had forgotten that we're not storing the filename.

> > If you both think we can get BMP/TIFF/PNG support in soonish, then I'll wait for that to happen. Otherwise let's wrap up these last bits and this is probably ready for initial release.
>
> Thoughts, Job?

I've raised a PR against this branch: colagrosso/se-web/pull/1. It adds support for BMP and PNG, but not TIFF because PHP has no builtin support and therefore implementation would take longer.

If it's accepted I think it's worth getting it into the initial release, as it includes database changes and we'd otherwise need to account for database migrations later on. TIFF support can probably wait for later.

acabal:
> I've raised a PR against this branch: [colagrosso/pull/1](https://github.com/colagrosso/se-web/pull/1). It adds support for BMP and PNG, but not TIFF because PHP has no builtin support and therefore implementation would take longer.

We can use `mime_content_type()` to verify that the file we got is actually a TIFF file, that's good enough.

acabal:
Converting TIFF to JPG thumbnails can be done using ImageMagick via `shell_exec('convert ...')`

You can convert directly into the destination file, no need to keep intermediary temp files around.

Make sure to check stderr to make sure nothing went wrong. It will be worth testing some bogus TIFFs (rename test.doc to test.tiff or whatever) to see what happens when ImageMagick has an error.

colagrosso:
Ok, Job blazed the trail and added support for PNG and BMP. Thanks, Job! I hacked in an ImageMagick convert for TIFF files to his framework. I tested it a few ways, and all the new image types work fine.

> Make sure to check stderr to make sure nothing went wrong. It will be worth testing some bogus TIFFs (rename test.doc to test.tiff or whatever) to see what happens when ImageMagick has an error.

Bogus files—regardless of extension—get caught by `mime_content_type()` before they reach ImageMagick. If `mime_content_type()` can't confirm the file is the correct type, the user sees this validation error:

Screenshot 2023-12-21 1 21 35 AM

Sorry this dragged on and got so close to the holidays (especially when folks are [asking about it](https://groups.google.com/g/standardebooks/c/lcwwsiKBxds) on the mailing list). Releasing now to trusted testers (like a few editors) sounds good to me, but if you'd like to wait to ensure that I'm around to respond to feedback, waiting until Jan 8, 2024 would be safe. I'm sure it will be a busy time for the site around Jan 1 with all the new PD releases.

colagrosso:
I added one small cleanup commit, and I'll be out until Jan 8, 2024. Happy holidays, both of you!

acabal:
OK, great work everyone! I'm going to merge this into a temp branch in this repo, and I'm going to do a little more work on it before we release. Hopefully I can get this released to the public first week of Jan.

acabal:
Alright, I've launched it! Thanks @colagrosso and @jobcurtis  for all of your hard work.

I decided to require a login for submitting artwork, instead of making it a public form. So I had to add password and some other logic to our login code.

Everything should be live now!

colagrosso:
Woohoo! It looks great. Thanks for managing the project, Alex, and for all the changes you added to get the database launched. 

I'll review the changes in d30ae9673cb6d94ae1c216088a09acf5e9efcb21 because there are a few things for me to learn from. Do you have any other feedback while it's fresh in your mind?

Last, you're welcome to close #234 and file new issues for any followup work you'd like. I have a few ideas, and I'm sure you do, too.

acabal:
Not especially, this was all around very good work. Most of the changes in that commit were for the password system which was out of scope here, and some code style tweaks. 

In general the most important point is validating every single thing the user sends in, because user input can and will be toyed with by script kiddies and various other bots in an attempt to cause havoc.

PR #293, opened 2023-10-31
Update wanted ebooks list

weijia-cheng:
Firefox complains that the *Circular Staircase* line is not properly formatted. I think the hyphens were throwing off the HTML parser.

I also removed Bierce (already done) and Hughes (well in progress) from the wanted list.


acabal:
Interesting, thanks. I fixed this myself before I saw this PR so I'll just close this

PR #295, opened 2023-11-05
Artist::DeleteIfUnused() is unused, so delete it

colagrosso:
b758ed4 removed the call to it. Also, with Artist::GetOrCreate() we're robust against creating duplicate artists, so an unused artist from a previous error will be reused when the user retries.


jobcurtis (approved):
Ironic :laughing:

PR #296, opened 2023-11-05
Replace string use with CONSTANT_VALUES

colagrosso:
Minor cleanup, Job.


colagrosso:
Also, I'm tacking on a second code cleanup commit here about `try/catch` formatting. I updated `if/else` in #282:

https://github.com/standardebooks/web/pull/282/commits/f2c7eb437d6a1f7cf56f470a0230bf87bb883d06

but I missed `try/catch` at the time.

jobcurtis (approved):

PR #297, opened 2023-11-11
Fix variable undefined errors found by PHPStan

colagrosso:


colagrosso:
Minor revision, merging without review.

PR #300, opened 2023-12-07
Update Docker Image to Ubuntu 22.04 and PHP 8.1

HgSlv:
- Updated base image to Ubuntu 22.04.
- Upgraded to PHP 8.1.

Resolves issues with composer dependencies previously incompatible with PHP 7.1.

### DIFF
--- a/vms/docker/start-server.sh
+++ b/vms/docker/start-server.sh
@@ -1,19 +1,43 @@
 #!/bin/sh
 
+# Generate SSL certificate if it doesn't exist
 if [ ! -f /standardebooks.org/web/config/ssl/standardebooks.test.crt ]; then
 	openssl req -x509 -nodes -days 99999 -newkey rsa:4096 -subj "/CN=standardebooks.test" -keyout /standardebooks.org/web/config/ssl/standardebooks.test.key -sha256 -out /standardebooks.org/web/config/ssl/standardebooks.test.crt
 fi
 
 cd /standardebooks.org/web
+
+# Update to required PHP version here if necessary
+
+# Run Composer install.
+# Note that the PHP version should be compatible with the packages in the composer.lock file.
 composer install
 
-ln -s /standardebooks.org/web/config/apache/standardebooks.test.conf /etc/apache2/sites-available/
-ln -s /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/*/cli/conf.d/
-ln -s /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/*/fpm/conf.d/
-ln -s /standardebooks.org/web/config/php/fpm/standardebooks.test.conf /etc/php/*/fpm/pool.d/
-a2ensite standardebooks.test
-service apache2 start
-service php7.4-fpm restart
+# Create symlink after checking if it already exists or removing the existing one
+create_symlink() {
+	target=$1
+	link=$2
+
+	if [ -L "$link" ]; then # Check if symlink already exists.
+		echo "Symlink $link already exists; skipping."
+	else
+		ln -s "$target" "$link"
+	fi
+}
+
+create_symlink /standardebooks.org/web/config/apache/standardebooks.test.conf /etc/apache2/sites-available/standardebooks.test.conf
+create_symlink /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/8.1/cli/conf.d/standardebooks.org.ini
+create_symlink /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/8.1/fpm/conf.d/standardebooks.org.ini
+create_symlink /standardebooks.org/web/config/php/fpm/standardebooks.test.conf /etc/php/8.1/fpm/pool.d/standardebooks.test.conf
+
+# Enable site only if it's not already enabled
+if ! a2query -s standardebooks.test > /dev/null; then
+	a2ensite standardebooks.test
+fi
+
+# Start or restart services
+service apache2 start || service apache2 reload
+service php8.1-fpm restart || service php8.1-fpm start
 
 # Keep the server available by holding open the container
 tail -f /dev/null
--- a/vms/docker/Dockerfile
+++ b/vms/docker/Dockerfile
@@ -1,23 +1,43 @@
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 
-RUN apt-get update
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y composer php-fpm php-cli php-gd php-xml php-apcu php-mbstring php-intl php-curl php-zip apache2 apache2-utils libfcgi0ldbl task-spooler libaprutil1-dbd-mysql attr libapache2-mod-xsendfile
-RUN DEBIAN_FRONTEND=noninteractive apt-get install -y sudo imagemagick openjdk-8-jre python3 pip calibre
-RUN apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+# Avoid user interaction with tzdata when installing packages
+ENV DEBIAN_FRONTEND=noninteractive
 
+# Install dependencies and required PHP extensions
+RUN apt-get update && \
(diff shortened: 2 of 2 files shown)


vms/docker/start-server.sh
```diff
@@ -1,19 +1,43 @@
...
+		ln -s "$target" "$link"
+	fi
+}
+
+create_symlink /standardebooks.org/web/config/apache/standardebooks.test.conf /etc/apache2/sites-available/standardebooks.test.conf
+create_symlink /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/8.1/cli/conf.d/standardebooks.org.ini
+create_symlink /standardebooks.org/web/config/php/fpm/standardebooks.org.ini /etc/php/8.1/fpm/conf.d/standardebooks.org.ini
+create_symlink /standardebooks.org/web/config/php/fpm/standardebooks.test.conf /etc/php/8.1/fpm/pool.d/standardebooks.test.conf
+
+# Enable site only if it's not already enabled
```
acabal, line 33: Do we really need to check if the site is already enabled, and if the services are already started? I thought this script ran once on container initialization. Does it actually run more than once?

HgSlv: Nope, it only runs once if all is good.
I had to run the script multiple times while troubleshooting, that's why i left it and [this one as well](https://github.com/standardebooks/web/pull/300/files#diff-b93d6ca6f7ccd12896996adfb1f7adfbefea8c59f3d5fbaec870e85515832a56R21).

acabal: OK, so let's leave those checks out, and also the checks on `service` and the `create_symlink` functino, because it's more code. We know this will all succeed because we're setting up the environment for that.

HgSlv: Oh, actually - i need to correct myself.
The `ENTRYPOINT` script isn't just for the first `docker run`. But also for every time the container is restarted with `docker start`.

> The `ENTRYPOINT` specifies a command that will always be executed when the container starts.

https://stackoverflow.com/a/34245657

HgSlv:
> OK, so is there a way to set Docker up so that this only runs exactly one time when the container is first initialized? We should put all of this code at that point, and not every time the system restarts.

[Mounting host volumes during build is not possible.](https://stackoverflow.com/a/26053710) 
That's why all the code that refers to these mounted files is on `start-server.sh`.

acabal:
But VOLUME doesn't appear in any of these scripts?

I don't know anything about Docker but I feel like I'm missing something here. Surely it must be possible to initialize a container once on creation, instead of re-initializing everything any time it restarts. What if I have a container with a database... does MySQL get uninstalled and reinstalled, and all of its data gets wiped, every time the container restarts?

HgSlv:
> But VOLUME doesn't appear in any of these scripts?
> 
> I don't know anything about Docker but I feel like I'm missing something here. Surely it must be possible to initialize a container once on creation, instead of re-initializing everything any time it restarts. What if I have a container with a database... does MySQL get uninstalled and reinstalled, and all of its data gets wiped, every time the container restarts?

After building the image we use `docker run -dp 443:443 -v "$(pwd):/standardebooks.org/web" standardebooks:latest`
-v: The flag used to create a bind mount volume.

HgSlv:
> That doesn't get to my question... there must be some way to have an image where the base state (like required packages, initial configuration, and initial data population) is all done exactly once, and not every time the container "restarts". Otherwise there is no such thing as "state" in a container and that doesn't make sense to me.
> 
> The changes you propose suggest to me that `start-server.sh` can run repeatedly and stomp on updates it previously made and attempt to start services that may already be started. This should not be what we're doing, but if it is, then we should rethink how we're doing Docker altogether, instead of adding checks to make sure symlinks don't already exist, or that a service has already started.
> 
> Or am I understanding Docker wrong? That is certainly possible. I'm expecting to behave like a lightweight app-level VM of sorts.

Yes, you could say Docker acts like a "Virtual machine" that shows the code changes, done on your host machine. 
Let's break it down. We use the following commands for docker:
> docker build . -t standardebooks -f vms/docker/Dockerfile

This command builds a Docker image named standardebooks using the `Dockerfile` found at vms/docker/Dockerfile.

> docker run -dp 443:443 -v "$(pwd):/standardebooks.org/web" standardebooks:latest

This command will start a new container from the `standardebooks:latest` image, mounting the current directory as a volume inside the container at /standardebooks.org/web. (At this point the `start-server.sh` gets run)


Now, the important part is: you can't use mounted volumes in the build (`Dockerfile`), that's why a lot of the scripting is in `start-server.sh`.

This has the consequence that, `start-server.sh` (specified as `ENTRYPOINT` on the `Dockerfile`) gets executed both when you first run the container with `docker run` and every time you restart it with `docker start`.

acabal:
According to [this](https://linuxhandbook.com/docker-run-vs-start-vs-create/), `docker run` combines `docker create` and `docker start`. But we don't want to create a new container every time we run it - we already created it once. I think this is where the problem is and my intuition was correct.

If I'm understanding this right, then we should use `docker start` instead of `docker run`. We're by calling `docker build` and then `docker run` we're creating the container twice, which is why problems appearing. So this appears to be a problem in the readme, not a problem in `start-server.sh`.

acabal:
Actually, now that I look at this closer, it looks like the readme is in fact correct. This is because it correctly builds the image, the container, and starts it with those two commands. But if you later stop the container, then you must start it with `docker start`, not `docker run`. Using `docker run` will create a new container even though we already created one. Does that sound right?

acabal:
Updated in fa312ea0

HgSlv:
Hey @acabal ! I saw that my PR got closed and noticed your new commit. Just curious, was there something off with my PR? I took a crack at explaining the code earlier, so I'm hoping it was helpful. A little acknowledgement of the effort can go a long way in maintaining a constructive open-source community. Cheers!

acabal:
Well, we don't want any of that extra check code because running `docker run` twice is user error. That's what I mentioned in the PR. We also don't want to change formatting because everyone changing formatting to their own style is just going to result in a lot of commit noise. We also don't need to specify specific PHP versions in those packages because the generic metapackages select the highest version anyway (8.1 in this case).

I saw you reacted to my last comment regarding `docker run` but didn't change the PR, so I just made the minor tweaks necessary for compatibility myself since I had a few spare minutes. However I don't know much about Docker and my intuition on the issues in this PR was correct so the rest I'm leaving to the person who wrote this code in the first place, so they can confirm it's working as intended.

HgSlv:
> Well, we don't want any of that extra check code because running `docker run` twice is user error. That's what I mentioned in the PR. We also don't want to change formatting because everyone changing formatting to their own style is just going to result in a lot of commit noise. We also don't need to specify specific PHP versions in those packages because the generic metapackages select the highest version anyway (8.1 in this case).
> 
> I saw you reacted to my last comment regarding `docker run` but didn't change the PR, so I just made the minor tweaks necessary for compatibility myself since I had a few spare minutes. However I don't know much about Docker and my intuition on the issues in this PR was correct so the rest I'm leaving to the person who wrote this code in the first place, so they can confirm it's working as intended.

No offense, but neither your "intuition" is correct, nor have you understood the implications [of what you wrote](https://github.com/standardebooks/web/pull/300#issuecomment-1847946791).
The proposed conditional changes concern running `docker start` multiple times; they are not about running `docker run` twice. [To reiterate](https://github.com/standardebooks/web/pull/300#discussion_r1420911701): 
> The ENTRYPOINT script (`start-server.sh`) isn't just run for the first `docker run`. It also runs every time the container is restarted with `docker start`.

Regrettably, the outcome of the commit is a broken docker run featuring a wrong PHP version and wildcards that do not work.

image

acabal:
OK. Well in any case I am not satisfied with the explanation given in this PR. I do not understand why VM-like software that inherently saves its state at some point would require that the state be recreated wholesale from scratch every time it is started. There absolutely must be a way to only initialize things exactly once, because otherwise how would software requiring data in a database function? That is not explained in this PR to my satisfaction.

I will fix the shell script issue, that is an issue with a shell wildcard and not with Docker.

Unfortunately I don't have the time to learn more about Docker right now and so this will have to wait for Robin to look at it.

PR #302, opened 2023-12-12
Update navigation footer script in Python

erinendrei:
deploy-ebook-to-www: 

- In order to make an instance of the SeEpub class in inject-chapter-navigation-footer we need the main directory of the ebook not workDir/src/epub.

inject-chapter-navigation-footer: 
- I couldn't run the script using pipx, so I had to run it using standardebooks installed with pip. The se package doesn't seem installable editably with pip, so I couldn't use any edits to tools I'd made; originally I did have a function in formatting.py as you suggested.

- A function for extracting titles from the spine could easily be added to formatting.py, but 1) I won't be able to test it, unless there is some way I'm missing to run the script with pipx or use a pip installation editably and 2) it's only one line here (line 26, setting value of rel_spine_paths) so maybe it's not worth adding anyway.

- I tested this by simply running it on a single ebook and inspecting the HTML output, which appears the same as that generated by the PHP script, only without the error for poetry collections.

- Currently the arguments cannot have trailing "/". This seems to conform to the current values of bookUrl and workDir in deploy-ebook-to-www.

- Here is the command I used to run it (obviously alter paths as required if also running it in this way):

`python3 web/scripts/inject-chapter-navigation-footer "ebooks-dir/robert-frost_new-hampshire" "https://standardebooks.org/ebooks/robert-frost/new-hampshire"`

### DIFF
--- a/scripts/deploy-ebook-to-www
+++ b/scripts/deploy-ebook-to-www
@@ -354,7 +354,7 @@ do
 		find "${workDir}"/src/epub \( -type d -name .git -prune \) -o -type f -name "*.xhtml" -print0 | xargs -0 sed --in-place --regexp-extended "s|<body([^>]*)>|<body\1><header><nav><ul><li><a href=\"/\">Standard Ebooks</a></li><li><a href=\"${bookUrl}\">Back to ebook</a></li><li><a href=\"${bookUrl}/text\">Table of contents</a></li></ul></nav></header>|"
 
 		# Add a chapter navigation footer to each page
-		"${scriptsDir}"/inject-chapter-navigation-footer "${workDir}"/src/epub "${bookUrl}"
+		"${scriptsDir}"/inject-chapter-navigation-footer "${workDir}" "${bookUrl}"
 
 		# Adjust sponsored links in the colophon
 		sed --in-place 's|<p><a href="http|<p><a rel="nofollow" href="http|g' "${workDir}"/src/epub/text/colophon.xhtml
--- a/scripts/inject-chapter-navigation-footer
+++ b/scripts/inject-chapter-navigation-footer
@@ -1,11 +1,11 @@
 #!/usr/bin/env python3
 
 import argparse
-import os
 import html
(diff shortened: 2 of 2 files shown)


acabal:
Thanks Erin, looks very good.

I ran it on the live version of [James Weldon Johnson's Poetry](https://standardebooks.org/ebooks/james-weldon-johnson/poetry/text) and there's a quirk: If you open the titlepage, the "next" footer item says `Imprint`. But it should say, `Poetry - Imprint`, and indeed that's what's in the imprint's `<title>` element. The rest of the files correctly list `Poetry - ` before each item. Can you check what's going on here?

acabal:
Can you also adjust the code style a little:
- double quotes `"` for all strings; strings containing double quotes should be triple-quoted `"""`
- Indent with 1 tab and not spaces

Thanks!

erinendrei:
Thanks Alex, the code style should be fixed now.

Is the output of my script still live for the Weldon Johnson book? Currently the titlepage's "Next:" link is `Poetry - Imprint` when I view it [here](https://standardebooks.org/ebooks/james-weldon-johnson/poetry/text/titlepage.xhtml).

acabal:
Hmm, yes, looks like it's working, so it must have been a cache issue on my end. The last thing is that footer links should not end in `.xhtml`, they should have no extension at all. Can you tweak that? In otherwords the titlepage should link to `https://standardebooks.org/ebooks/james-weldon-johnson/poetry/text/imprint` (note no extension), etc.

erinendrei:
OK, I believe that's fixed now.

erinendrei:
Sorry, I've messed up here. I was simply trying to reword three commit messages.  Should I delete this and make a new PR with only the relevant commits?

erinendrei:
Force-pushing the commits from a new branch that was up to date with master seems to have corrected it.

acabal:
OK great, thanks! I'll run this soon and the corpus should update over the next few days.

PR #304, opened 2023-12-24
Add how-to poem guide

EmmaSweeney:


acabal:
Great work Emma. I think this would benefit from a few screenshots so people can see what the styling actually looks like. Can you put that together? You can put them in the `www/images` folder, make sure to use both avif and jpg plus @2x variants (see other images in that folder for an example of what I mean).

EmmaSweeney:
I knew I was forgetting something. I'll have this done in the next couple of days.

EmmaSweeney:
Ok, this has pictures now.

acabal:
Great work, thanks!

PR #307, opened 2024-01-01
Add check for rsvg-convert

sweetgiorni:
`scripts/deploy-ebook-to-www` relies on `rsvg-convert` which doesn't seem to be a documented dependency:

```sh
➜  web git:(master) ./scripts/deploy-ebook-to-www ../ebooks/david-lindsay_a-voyage-to-arcturus.git
./scripts/deploy-ebook-to-www: line 250: rsvg-convert: command not found
```

This PR adds a check for it and suggests installing `python3-sphinxcontrib.svg2pdfconverter` to provide it.

### DIFF
--- a/scripts/deploy-ebook-to-www
+++ b/scripts/deploy-ebook-to-www
@@ -130,6 +130,7 @@ fi
 require "convert" "Try: apt-get install imagemagick"
 require "git" "Try: apt-get install git"
 require "php" "Try: apt-get install php-cli"
+require "rsvg-convert" "Try: apt-get install librsvg2-bin"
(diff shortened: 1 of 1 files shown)


acabal:
Thanks, however the correct package is `librsvg2-bin`, can you update that?

sweetgiorni:
> Thanks, however the correct package is `librsvg2-bin`, can you update that?

Weird, `python3-sphinxcontrib.svg2pdfconverter` is the only result from `apt search rsvg-convert`... 🤷 Fixed.

acabal:
Thanks!

PR #310, opened 2024-01-02
Update wanted ebooks list

btaylor8:
I noticed there are a few books on the wanted list that have already been released.

### Removed

- Early Autumn, by Louis Bromfield
- Trilby, by George du Maurier
- The House of the Wolflings (sic), by William Morris


acabal:
Thanks!

PR #316, opened 2024-01-10
Allow editing artwork in covers database

colagrosso:
This isn't quite ready yet, but putting a draft up here. I'll discuss a bit in the issue.

### DIFF
--- a/config/apache/standardebooks.org.conf
+++ b/config/apache/standardebooks.org.conf
@@ -302,6 +302,8 @@ Define webroot /standardebooks.org/web
 	RewriteCond expr "tolower(%{REQUEST_METHOD}) =~ /^post$/"
 	RewriteRule				^/artworks/([^/\.]+)/([^/\.]+)$			/artworks/post.php?artist-url-name=$1&artwork-url-name=$2 [L]
 
+	RewriteRule				^/artworks/([^/\.]+)/([^/\.]+)/edit$		/artworks/edit.php?artist-url-name=$1&artwork-url-name=$2 [L]
+
 	# Specific config for /bulk-downloads
 	<DirectoryMatch "${webroot}/www/bulk-downloads">
 		# Both directives are required
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -690,8 +690,6 @@ class Artwork extends PropertiesBase{
 
 		$this->Validate($uploadedFile);
 
-		$this->Updated = new DateTime();
-
 		$tags = [];
 		foreach($this->Tags as $artworkTag){
 			$tags[] = ArtworkTag::GetOrCreate($artworkTag);
@@ -709,7 +707,6 @@ class Artwork extends PropertiesBase{
 			CompletedYear = ?,
 			CompletedYearIsCirca = ?,
 			Created = ?,
-			Updated = ?,
 			Status = ?,
 			SubmitterUserId = ?,
 			ReviewerUserId = ?,
@@ -726,7 +723,7 @@ class Artwork extends PropertiesBase{
 			where
 			ArtworkId = ?
 		', [$this->Artist->ArtistId, $this->Name, $this->UrlName, $this->CompletedYear, $this->CompletedYearIsCirca,
-				$this->Created, $this->Updated, $this->Status, $this->SubmitterUserId, $this->ReviewerUserId, $this->MuseumUrl, $this->PublicationYear, $this->PublicationYearPageUrl,
+				$this->Created, $this->Status, $this->SubmitterUserId, $this->ReviewerUserId, $this->MuseumUrl, $this->PublicationYear, $this->PublicationYearPageUrl,
 				$this->CopyrightPageUrl, $this->ArtworkPageUrl, $this->IsPublishedInUs, $this->EbookWwwFilesystemPath, $this->MimeType, $this->Exception, $this->Notes,
 				$this->ArtworkId]
 		);
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -20,6 +20,9 @@ use function Safe\preg_replace;
  * @property string $ImageUrl
  * @property string $ThumbUrl
  * @property string $Thumb2xUrl
+ * @property string $ImageFsPath
+ * @property string $ThumbFsPath
+ * @property string $Thumb2xFsPath
  * @property string $Dimensions
  * @property ArtworkStatus|string|null $Status
  * @property Ebook $Ebook
@@ -55,6 +58,9 @@ class Artwork extends PropertiesBase{
 	protected ?string $_ImageUrl = null;
 	protected ?string $_ThumbUrl = null;
 	protected ?string $_Thumb2xUrl = null;
+	protected ?string $_ImageFsPath = null;
+	protected ?string $_ThumbFsPath = null;
+	protected ?string $_Thumb2xFsPath = null;
 	protected ?string $_Dimensions = null;
 	protected ?Ebook $_Ebook = null;
 	protected ?Museum $_Museum = null;
@@ -209,7 +215,7 @@ class Artwork extends PropertiesBase{
 				throw new Exceptions\InvalidArtworkException();
 			}
 
-			$this->_ImageUrl = COVER_ART_UPLOAD_PATH . $this->ArtworkId . $this->MimeType->GetFileExtension();
+			$this->_ImageUrl = COVER_ART_UPLOAD_PATH . $this->ArtworkId . $this->MimeType->GetFileExtension() . '?ts=' . $this->Updated?->getTimestamp();
 		}
 
 		return $this->_ImageUrl;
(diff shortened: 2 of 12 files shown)


www/artworks/edit.php
```diff
@@ -0,0 +1,232 @@
+<?
+use function Safe\gmdate;
```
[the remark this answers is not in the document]
colagrosso: Done. Thanks for all the updates to `artworks/new.php`. I was able to reuse all of them here via template.

www/artworks/index.php
```diff
@@ -4,14 +4,16 @@
...
-$status = HttpInput::Str(GET, 'status', false) ?? null;
+$status = HttpInput::Str(GET, 'status', false) ?? COVER_ARTWORK_STATUS_APPROVED;
 $sort = HttpInput::Str(GET, 'sort', false);
 $pages = 0;
 $totalArtworkCount = 0;
 $pageDescription = '';
 $pageTitle = '';
 $queryString = '';
 $isAdminView = $GLOBALS['User']?->Benefits?->CanReviewArtwork ?? false;
+$submitterUserId = $GLOBALS['User']->UserId ?? null;
```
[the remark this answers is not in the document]
colagrosso: I need a user ID to pass to `Library::FilterArtworks()` below. `$submitterUserId` seems like a natural name, but I'm open to changing it. (Or a bigger change to the `FilterArtworks` API. I will still rewrite it with pure SQL and a dedicated column to index and search on.)

www/css/artwork.css
```diff
@@ -69,13 +69,13 @@ ol.artwork-list > li img{
 	max-width: 100%;
 }
 
-form[action^="/artworks/"]{
+form:not(.artwork-operate){
```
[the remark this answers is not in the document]
colagrosso: This was some bad CSS. Sorry about that. I tried harder and was able to avoid adding a class. But there are some complicated selectors because there are 4 `<form>`s that have very similar `action` and `method` attributes:

1. The submission form in `artworks/new.php`
2. The edit form in `artworks/edit.php`
3. The Editor Options form in `artworks/get.php`
4. The browse from in `artworks/index.php`

I'm clearly terrible at CSS, including when to decide to give up and add a class to a form or two. Let me know how these selectors look to you.

config/apache/standardebooks.org.conf
```diff
@@ -302,6 +302,9 @@ Define webroot /standardebooks.org/web
 	RewriteCond expr "tolower(%{REQUEST_METHOD}) =~ /^post$/"
 	RewriteRule				^/artworks/([^/\.]+)/([^/\.]+)$			/artworks/post.php?artist-url-name=$1&artwork-url-name=$2 [L]
```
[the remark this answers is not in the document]
colagrosso: Done, good call.

lib/Artwork.php
```diff
@@ -618,8 +623,19 @@ public function Create(array $uploadedFile): void{
...
 		$this->Validate();
 
+		// Can't assign directly to $this->Tags because it's hidden behind a getter
+		$tags = [];
+		foreach($this->Tags as $artworkTag){
+			$tags[] = ArtworkTag::GetOrCreate($artworkTag);
+		}
+		$this->Tags = $tags;
+
+		$this->Artist = Artist::GetOrCreate($this->Artist);
```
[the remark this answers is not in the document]
colagrosso: I believe `GetOrCreate()` is doing the correct thing here, modulo a few edge cases.

1. If the user updates the artist from `John Nobody` to `John Q. Nobody`, then yes we will create a new record and the old one will be a zombie. It would have been better to update the `Artist` record. But if the user updates the artist from `Claude Monet` to `John Q. Nobody`, we don't want to update `Artist` record. At this point in the code, it's hard to tell those two cases apart because we've already lost the old `Artist`. If the user updates the artist from `Claude Monet` to `John Singer Sargent`, `Artist::GetOrCreate()` works as expected and looks up the existing artist.

2. It's not possible to edit just the artist's death year on that form. This is consistent with the advice on the form, "_For existing artists, leave the year of death blank._"

Let me know if you'd like me to try harder here.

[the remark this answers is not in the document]
acabal: OK. So what we should do is after updating the artist, delete any artists in the DB who are not attached to an artwork.

www/artworks/index.php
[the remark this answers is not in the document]
acabal: OK

www/css/artwork.css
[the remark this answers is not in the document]
acabal: It's OK to use classes if it helps in readability. We want selectors to be basically readable and once you get into multiple nth-child territory it's not as readable anymore. The approach to style for ebooks and web apps is different. The key is not to use special class for literally everything. Classes can be used for clarity but they can also be smartly organized.

lib/Artwork.php
```diff
@@ -665,6 +688,7 @@ public function Save(): void{
 			CompletedYear = ?,
 			CompletedYearIsCirca = ?,
 			Created = ?,
+			Updated = ?,
```
acabal, line 691: We don't have to set the Updated column because it's of type `timestamp`, which the DB automatically updates any time the row changes.

lib/Artwork.php
```diff
@@ -681,10 +705,29 @@ public function Save(): void{
...
+
+		foreach($this->Tags as $tag){
+			Db::Query('
+				INSERT INTO ArtworkTags (ArtworkId, TagId)
+				VALUES (?,
+				        ?)
+			', [$this->ArtworkId, $tag->TagId]);
+		}
+
+		if(!empty($uploadedFile) && $uploadedFile['error'] == UPLOAD_ERR_OK){
```
acabal, line 728: By the time we get here, we should already know that the upload was successful. If it wasn't then it should have thrown an exception in `Validate()`. So no need to check for error status here.

lib/Library.php
```diff
@@ -180,6 +182,19 @@ public static function FilterArtwork(string $query = null, string $status = null
 				SELECT *
 				from Artworks', [], 'Artwork');
 		}
+		elseif($status == 'all-submitter' && $submitterUserId !== null){
+			$artworks = Db::Query('
+				SELECT *
+				from Artworks
+				where Status in ("approved", "in_use")
+				or (Status = "unverified" and SubmitterUserId = ?)', [$submitterUserId], 'Artwork');
```
acabal, line 190: Don't bake in `unverified` as as a string, use the enums for this

lib/Library.php
```diff
@@ -180,6 +182,19 @@ public static function FilterArtwork(string $query = null, string $status = null
...
+		}
+		elseif($status == 'unverified-submitter' && $submitterUserId !== null){
+			$artworks = Db::Query('
+				SELECT *
+				from Artworks
+				where Status = "unverified" and SubmitterUserId = ?', [$submitterUserId], 'Artwork');
```
acabal, line 196: Same

www/artworks/edit.php
```diff
@@ -0,0 +1,73 @@
+<?
+use function Safe\session_unset;
+
+session_start();
+
+$exception = $_SESSION['exception'] ?? null;
+
+try{
+	$artwork = Artwork::GetByUrl(HttpInput::Str(GET, 'artist-url-name') ?? '', HttpInput::Str(GET, 'artwork-url-name') ?? '');
```
acabal, line 9: We have to also pull the artwork from the session, not just the URL. If the user submitted edits but they were rejected, the form must show they changes they made, not the original artwork. This is the same logic in the new artwork form, you can basically copy and paste that logic into here for the most part.

www/artworks/post.php
```diff
@@ -139,6 +176,9 @@
 	if($httpMethod == HTTP_PATCH && $artwork !== null){
 		header('Location: ' . $artwork->Url);
 	}
+	else if($httpMethod == HTTP_PUT && $artwork !== null){
```
acabal, line 179: After PUTting we want to redirect to the new artwork, not the edit form. Logic should be the same as PATCH

acabal (changes requested):
Good work, some minor notes

lib/Artwork.php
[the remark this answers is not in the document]
colagrosso: Yep, good call, done. It was a straightforward implementation, too.

lib/Artwork.php
colagrosso: Thanks for explaining that. I also see it now in the schema. Removed.

lib/Artwork.php
colagrosso: Since the upload is optional during `Save()` these are the checks necessary to see if the user uploaded a valid file during their edit. I agree it looks like validation logic, so I added a comment here. I believe the check is necessary. I can elaborate more if the comment still isn't clear.

lib/Library.php
colagrosso: Done. I'll try to clean this up more in the next pass through `Library::FilterArtwork()`.

lib/Library.php
colagrosso: Done.

www/artworks/edit.php
colagrosso: Yep, done.

www/artworks/post.php
colagrosso: That already works, i.e., the user does get redirected to the new artwork after a successful PUT. That happens above on line 105. This is the code that redirects the user after an error. I believe it's doing the right thing. 

(I believe the PR was confusing because Github was hiding so many unchanged lines between lines 105 and 179.)

colagrosso:
I forgot about simplifying the form CSS, so let me take one more pass at that before you re-review, please.

colagrosso:
Ok, thanks for your patience. Ready for another review.

acabal:
OK, looks very good, thanks! I'm going to make a few small tweaks and hopefully release this later today or tomorrow. Great work!

PR #317, opened 2024-01-10
Add grid layouts for smaller screens to /artwork grid

jobcurtis:
With the new grid it was a bit difficult to view the /artworks page on mobile. The filter controls still overflow but it is usable.

fyi @colagrosso in case you are working in this area.

screenshot-1
screenshot-2

### DIFF
--- a/www/css/artwork.css
+++ b/www/css/artwork.css
@@ -276,3 +276,23 @@ form div.footer{
 		grid-column: 1;
 	}
 }
+
+@media(max-width: 1020px){
+	main > section > ol.artwork-list{
+		grid-template-columns: repeat(3, 1fr);
+		gap: 3rem;
+	}
+}
+
+@media(max-width: 760px){
+	main > section > ol.artwork-list{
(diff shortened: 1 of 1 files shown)


acabal:
Good, thanks, this was on the to do list. 

Does this also prevent the search bar from creating a horizontal scrollbar on mobile? That's also we want to fix - the `/artworks` is not usable on mobile at all right now.

jobcurtis:
No, I haven't touched the filter controls. This is just low-hanging fruit to make the artwork actually visible on mobile. With this PR, there's still a horizontal scroll bar but the page is usable. 

I can take a look at fixing the filter controls tomorrow, either on this PR or a new one.

acabal:
OK, thanks! Yes, if you could work on the search bar so the entire page is properly responsive, that would be great.

See how the ebooks search bar is done. It's done with CSS Grid which is convenient for arranging bars like this in an attractive grid pattern on small screens.

PR #318, opened 2024-01-11
Add grid layouts for smaller screens to /artwork filter form

jobcurtis:
As discussed, updates to the filter controls on `/artworks` for smaller screens. Screenshots below;

screenshot-fullscreen
screenshot-tablet
screenshot-phone


acabal:
Great, thanks. A minor nitpick: Can you make the Filter button the same height as the per page combo box? Right now it's slightly smaller and it looks off. See the ebook search form to see how that's done with a top margin.

jobcurtis:
Sure, I've fixed the button height. That `align-items: end` was causing the different behaviour, and not achieving anything as far as I can tell.

acabal:
Excellent, thanks!

PR #319, opened 2024-01-15
Default status is Unverified when posting artwork

colagrosso:
Without this change, users that have `CanUploadArtwork` but not `CanReviewOwnArtwork` would get an `InvalidPermissionsException` because `Status` is `null` on this line:

https://github.com/standardebooks/web/blob/206e125da79357adc08d9e54857aceb880b0b829/www/artworks/post.php#L48

Before 531e3600eadd3d9e79d5f2b4b19d762ba60ec40a, the line read:

```php
$artwork->Status = HttpInput::Str(POST, 'artwork-status', false, COVER_ARTWORK_STATUS_UNVERIFIED);
```


acabal:
Thanks!

PR #320, opened 2024-01-16
Remove unused Artist::FindMatch 

colagrosso:


acabal:
Ah, I end up reading PRs after my regular inbox, and someone already reported the error, and I fixed this in an earlier commit. I can accept removing the unused method, can you edit this PR to just include that? Or you can include that when the edit functionality is done.

I think I'm done with updates to the artwork system for now, so unless there are bug fixes on my end, hopefully you won't have to do any annoying pull merges until the edit functionality is good to go.

colagrosso:
Sure, no prob. It was useful to see how you fixed the query in 56ebdcab0fa94c3cdeb1b7e9f5ceac31fe0e0620. 

You made so many improvements to the artwork system in such a short about of time. Way to go! I'm impressed. 

Someday you'll have to tell me how you got all the PD proof entered for the hundreds of existing artworks. My obsolete script didn't handle that.

The edit functionality is almost good to go. After that, I've noticed a few other things along the way, plus there's the todo list in the README. I'll keep going on the artwork system even though I'm slower than you.

acabal:
Thanks!

PR #321, opened 2024-01-19
Library::FilterArtwork(): Move sort logic to SQL, condense status condition

colagrosso:
This is just a preliminary refactor because I haven't added the filter logic to the SQL yet, but I wanted to check if you thought I was on the right track with the sort logic and handling the status. The goal is to have just one `Db::Query()` instead of one in each of the `if`/`elseif` branches because the SQL is about to get gnarlier.

### DIFF
--- a/lib/Library.php
+++ b/lib/Library.php
@@ -180,9 +180,6 @@ class Library{
 		elseif($status == 'all-admin'){
 			$statusCondition = 'true';
 		}
-		elseif($status == 'in-use'){
-			$statusCondition = 'EbookWwwFilesystemPath is not null';
-		}
 		elseif($status == 'all-submitter' && $submitterUserId !== null){
 			$statusCondition = 'Status = ? or (Status = ? and SubmitterUserId = ?)';
 			$params[] = ArtworkStatus::Approved->value;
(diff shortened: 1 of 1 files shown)


lib/Library.php
```diff
@@ -170,58 +170,56 @@ public static function FilterArtwork(string $query = null, string $status = null
...
-				from Artworks', [], 'Artwork');
+			$statusCondition = 'true';
 		}
 		elseif($status == 'in-use'){
-			$artworks = Db::Query('
-				SELECT *
-				from Artworks
-				where EbookWwwFilesystemPath is not null
-				', [], 'Artwork');
+			$statusCondition = 'EbookWwwFilesystemPath is not null';
```
acabal, line 184: This matches the previous logic but I think there was an error in the previous logic. We should also check that the status is 'approved' so we don't accidentally show declined artwork that has been marked as in use.

acabal (commented):
Looks good. Minor note logic error that is present in master.

lib/Library.php
colagrosso: You're right. I confirmed that was a problem before and fixed it here.

If you think this PR is good to go, you can squash these 3 little commits and merge it. I will have a whack at the upgraded search functionality soon.

acabal:
Great work, thanks!

PR #322, opened 2024-01-19
Fix typo when checking artwork upload image size

jobcurtis:
I think this is what is causing the upload exceptions mentioned on the mailing list. It's currently throwing an exception if the upload size is greater than 0 bytes.


colagrosso:
The fix looks right and it corrects the upload error for me locally.

PR #324, opened 2024-01-21
Keep words with apostrophes intact when searching

colagrosso:
The code will convert a search for:

`Whistler's Mother`

into:

`\b(whistler's|mother)\b`

Before this fix, it was:

`\b(whistler|s|mother)\b`

### DIFF
--- a/lib/Library.php
+++ b/lib/Library.php
@@ -213,10 +213,14 @@ class Library{
 			$orderBy = 'art.CompletedYear desc';
 		}
 
-		// Remove diacritics and non-alphanumeric characters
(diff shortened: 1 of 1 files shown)


colagrosso:
Sorry about the mistake before and thanks for noticing the problem. I targeted the fix to words with apostrophes, which might be too narrow, but I can't think of other searches with non-alphanumeric characters that we'd like to support.

Let me know if I should add more documentation here or even put this logic in a function with some PHPUnit tests.

acabal:
I think this works, thanks!

PR #326, opened 2024-01-23
List artwork by artist via /artworks/<artist-name>

colagrosso:
Let me know if I should leave the README.md file alone.

I'm sure you know this, but you'll want to run:

```sql
 CREATE INDEX index3 ON Artworks (ArtistId);
```

I confirmed that the explain plan uses the `UrlName` index on the `Artists` table (`idxUnique`) and then this new index on the `Artworks` table.


acabal:
Excellent work, thanks!

PR #327, opened 2024-01-23
Add the drop down arrows "Sort" and "Per page" elements on the artworks search form

colagrosso:
Minor changes that were bugging me

### DIFF
--- a/www/css/core.css
+++ b/www/css/core.css
@@ -1818,6 +1818,7 @@ label.select > span + span::after{
 	margin-top: -3px; /* Adjust for Crimson Pro line-height */
 	text-shadow: 1px 1px 0 rgba(255, 255, 255, .5);
 	transition: color .5s;
+	z-index: -1;
 }
 
 label.search:focus-within::before,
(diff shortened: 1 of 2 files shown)


www/css/core.css
```diff
@@ -1818,6 +1818,7 @@ label.select > span + span::after{
 	margin-top: -3px; /* Adjust for Crimson Pro line-height */
 	text-shadow: 1px 1px 0 rgba(255, 255, 255, .5);
 	transition: color .5s;
+	z-index: -1;
```
acabal, line 1821: Why is the z-index changed?

colagrosso: Without this change clicking exactly on the down arrow eats the click and doesn't activate the `<select>` element.

You can reproduce this today on:

https://standardebooks.org/ebooks
https://standardebooks.org/artworks

by clicking directly on the down arrow:

Screenshot 2024-01-23 at 1 30 15 PM

acabal:
Ah, OK, good catch! Thanks!

PR #328, opened 2024-01-23
Update punctuation outside quotation marks command in review guide

weijia-cheng:
It seems the command we currently have in the review guide is different than the one in the producer guide. It doesn't check for double quotes, which seems like an oversight. It makes sense to me to use the same command as the producers would use to check.


acabal:
Thanks!

PR #329, opened 2024-01-24
Add status and user filtering to GetArtworksByArtist

colagrosso:
I tested this for the reviewer, submitter, and non-logged in cases over a mix approved and unverified artwork.

### DIFF
--- a/lib/Library.php
+++ b/lib/Library.php
@@ -182,7 +182,7 @@ class Library{
 			$statusCondition = 'true';
 		}
 		elseif($status == 'all-submitter' && $submitterUserId !== null){
-			$statusCondition = 'Status = ? or (Status = ? and SubmitterUserId = ?)';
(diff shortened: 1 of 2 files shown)


lib/Library.php
```diff
@@ -182,7 +182,7 @@ public static function FilterArtwork(string $query = null, string $status = null
 			$statusCondition = 'true';
 		}
 		elseif($status == 'all-submitter' && $submitterUserId !== null){
-			$statusCondition = 'Status = ? or (Status = ? and SubmitterUserId = ?)';
+			$statusCondition = '(Status = ? or (Status = ? and SubmitterUserId = ?))';
```
colagrosso, line 185: These parentheses aren't necessary, but I added them to be consistent with the line I added below (where they are necessary).

acabal:
Excellent, thanks!

PR #330, opened 2024-01-25
Set the death year to null for the Anonymous artist

colagrosso:
It's weird that pages like this:

https://standardebooks.org/artworks/anonymous/still-life-with-artichokes-and-a-parrot

currently show:

| Artist | [Anonymous](https://standardebooks.org/artworks/anonymous) (d. 1850) |
| --- | ----------- |

You're welcome to just set the death year back to NULL in the DB—I don't feel too strongly about this one—but I thought this would be more robust to future reviewers unintentionally setting it again.

### DIFF
--- a/lib/Artist.php
+++ b/lib/Artist.php
@@ -3,6 +3,7 @@ use Safe\DateTime;
 use function Safe\date;
 
 /**
+ * @property ?int $DeathYear
  * @property string $UrlName
  * @property string $Url
  * @property array<string> $AlternateNames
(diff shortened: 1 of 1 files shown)


acabal:
Good point, but I think it should be the other way around: we should force storage of the correct value (`null`) and not amend  the results. So you can do this with a setter instead of a getter. You can also do a check in `Artist::Validate()` but instead of throwing an error you can just set it to `null` quietly.

acabal:
Storing it in the DB correctly allows us to run correct queries outside of the app, should we ever need a list of artworks with artists of unknown death year for example.

colagrosso:
Even better, thanks!

You'll have to do one last manual update to the record:

```sql
UPDATE Artists SET DeathYear = NULL WHERE UrlName = 'anonymous';
```

But from then on, the code won't allow a non-null death year for Anonymous. The reason that you can't just go to the `/edit` form and clear out the death year for Anonymous (like you can for other artists) is that the setter is invoked on object creation so that `$newDeathYear` and `$this->Artist->DeathYear` are already both `null` on this line:

https://github.com/standardebooks/web/blob/fc42f1579ae5083ece09179a04834ff60f0c0aed/lib/Artwork.php#L763

so the `UPDATE` is skipped for Anonymous.

acabal:
OK great, thanks!

PR #333, opened 2024-01-30
Conver `SORT_*` and `SOURCE_*` constants to enums

colagrosso:
These were minor changes I had in progress before taking on the ebooks DB that I wanted to finish


colagrosso:
Probably obvious, but I had to clear and rebuild the cache after this change. Without the rebuild I would get `apcu_fetch()` offset errors because of the mismatch between the `int` and `EbookSourceType` enum.

PR #340, opened 2024-02-09
Update the ebook URL validation error message

colagrosso:
I saw the [feedback](https://groups.google.com/g/standardebooks/c/8GOMcTJp_cI/m/NeJJH0ihAAAJ) on the mailing list, and even though this PR doesn't totally address it, the extra text might help future submitters sort out the problem on their own.

### DIFF
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -507,12 +507,12 @@ class Artwork extends PropertiesBase{
 			$error->Add(new Exceptions\MissingPdProofException());
 		}
 
-		// Check the ebook www filesystem path.
+		// Check the ebook URL.
 		// We don't check if it exists, because the book might not be published yet.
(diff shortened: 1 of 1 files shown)


acabal:
That message is not entirely correct because there might be a translator/illustrator as well. Instead we could suggest that the submitter use the URL in the dc:identifier which is exactly what we want anyway.

colagrosso:
Yep, done. 

Back to that one feedback report, in the manager's defense the [repo](https://github.com/BR903/mary-wollstonecraft_a-vindication-of-the-rights-of-woman) in the [thread](https://groups.google.com/g/standardebooks/c/ku7gWKz0KDI) was empty (as of Feb 10), so there was no `dc:identifier` to work from. Given this is rare, I don't think this will be a problem in the future.

acabal:
Thanks!

PR #342, opened 2024-02-16
fix a link redirect

partev:
https://docs.microsoft.com/en-us/windows/wsl/install-win10 -> https://learn.microsoft.com/en-us/windows/wsl/install


acabal:
Thanks!

PR #347, opened 2024-02-25
Clarify non-CC0 museum grandfather exceptions

weijia-cheng:
It's always kind of bugged me that we have two exceptions to the CC0 rule and it creates the impression that we are inconsistent about our own rules when people notice this. It has definitely come up on the mailing list before. In this PR I suggest an explicit explanation of why these museums are on the list, which also rules out any future exceptions. I hope I represented the situation correctly.

### DIFF
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -303,16 +303,6 @@ class Artwork extends Accessor{
 			return true;
 		}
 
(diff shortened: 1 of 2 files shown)


acabal:
How common is it to have a museum use the actual CC-PD mark instead of a CC0 license or a generic "public domain" text? Maybe it's simply time to accept the CC-PD mark in general.

PR #352, opened 2024-04-24
First stab at an accessibility page

robinwhittleton:
The goal here is to keep it concise: mission, ebooks, web and support. We can adapt this over time as necessary, but this feels good enough as a first attempt.

### DIFF
--- a/templates/Footer.php
+++ b/templates/Footer.php
@@ -42,6 +42,9 @@
 			<li>
 				<a href="/feeds">Ebook Feeds</a>
 			</li>
+			<li>
+				<a href="/about/accessibility">Accessibility</a>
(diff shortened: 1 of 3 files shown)


acabal:
Looks pretty good. The only thing to note is that in the header template, `highlight` refers to which top nav item is currently highlighted. There is no item for `accessibility`, so we should either highlight `about` or nothing.

acabal:
Should this live under `/about`? I.e. `/about/accessibility`?

robinwhittleton:
Yeah, that feels reasonable. I’ll fix that here, and update the ONIX templates too.

acabal:
Great, thanks!

PR #356, opened 2024-05-04
Add guidelines for handling compound words

lukasbystricky:
I added the two cases I've been told not to modernize. Are there any others?


acabal:
Great, thanks.

I'll merge this in, but I think this section is getting a little big for the step by step guide. Maybe we should create a new page to list our modernization rules in complete detail, and then link to there from the step by step guide instead. Would you be interested in working on that?

lukasbystricky:
Sure. Would it be more than a cut-paste job?

acabal:
Maybe not for now, though I can imagine we could certainly fill out a page like that in the future.

PR #358, opened 2024-05-09
Add new page for proofreading guidelines

lukasbystricky:
This is the new page for proofreading, with some added text on compound words.


acabal:
Great, thanks!

PR #365, opened 2024-05-24
Workaround a bug in Safe-PHP that causes PHPStan errors

colagrosso:
Thanks for pointing this out in a [previous commit](https://github.com/standardebooks/web/commit/406eff56a904b8d3b5fd1ad6a55fae860e5e1330#r142337290) and sorry for not noticing the inline PHPDocs you already added to previous `Safe\DateTimeImmutable` instantiations. 

I added a second commit in this PR to remove some leading slashes I mistakenly put in some `@throws` statements.

### DIFF
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -581,7 +581,7 @@ class Ebook{
 	// *******
 
 	/**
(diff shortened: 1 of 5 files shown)


acabal:
Thanks!

Issue #336, opened 2024-02-04
Convert ebooks storage to the database


colagrosso:
This is a large project that may require multiple issues, but we'll start here.

To help with the discussion, I have a draft PR at #335 where I hacked up a schema, script, methods, etc. for `Ebooks` and `EbookTags`. No hard feelings if we have to scrap that PR and start a different way.

Question:
* Would you be willing to create a separate, long-running branch like we did for `covers`? I think that would save you review time because you could review the changes periodically, and there would be no chance of breaking something in `master`.

Some notes about draft PR #335:
* For now the script just reuses the `Ebook` constructor. Eventually we will have to break up that constructor into separate methods. Some of the logic is needed for populating the DB tables, and some is needed for setting request-time properties.
* I haven't modified `Library::FilterEbooks()`, `/ebooks/index.php`, or `/ebooks/ebook.php` yet. I thought I would populate more fields in the DB tables first.
* We'll need some other tables for Contributors, Git commits, Collections, etc.

The more I look, the more work I see to do, but we'll chip away at it. One nice thing is that the ebook webpages are all read-only (unlike the artwork site).

colagrosso:
I forgot to mention above, but I ran the script in #335 over a hundred or so ebooks, and it picked up the title, URL, and tags just fine. You can skip rebuilding the books and just update the database via:

```bash
for BOOK in $(find /standardebooks.org/ebooks -maxdepth 1 -type d)
do
  tsp nice /standardebooks.org/web/scripts/deploy-ebook-to-www -v --no-build --no-images --no-epubcheck --no-recompose --no-feeds --no-bulk-downloads --update-ebook-database $BOOK
done
```

acabal:
I think we will probably drop `Ebook::__construct()` entirely in favor of several `::From*` methods, like `Ebook::FromFilesystem($path)`, `Ebook::Get($ebookId)`, etc. much like `Artwork`.

colagrosso:
Alex, the `db-rewrite` branch is ready for you to test drive.

The `db-rewrite` branch will merge cleanly with `master`, and you can preview the changes here: https://github.com/standardebooks/web/compare/db-rewrite?expand=1

If you have a working site from the `master` branch, you'll need to add these tables:

```
config/sql/se/CollectionEbooks.sql  
config/sql/se/Collections.sql  
config/sql/se/Contributors.sql  
config/sql/se/EbookLocSubjects.sql  
config/sql/se/EbookSources.sql  
[...]

PR #366, opened 2024-05-24
Complete the DB write path

colagrosso:
There are 5 commits in this PR, 3 small, 1 large, and 1 extra large. I could send these commits one at a time in different PRs if you would prefer. I don't think breaking up the extra large commit would make it clearer because it's attacking the large constructor.

1. (extra large) Rename the constructor `Ebook::__construct()` to `static Ebook::FromFilesystem()`

We [discussed](https://github.com/standardebooks/web/issues/336#issuecomment-1928001431) this back in February in #336. 

I added `GetFoo()` methods for all the derived properties like `GetUrl()`, `GetHasDownloads()`, etc. That logic is reusable to `Ebook` objects no matter if they were created from `Ebook::FromFilesystem()` or `Ebook::GetByIdentifier()`.

The result is that `Ebook.php` got more verbose. I personally think the logic for each property is now clearer, but it might be a shock to see how much bigger the file is.

Also in the commit: `Collection`, `Contributor`, `EbookSource`, and `GitCommit` need zero-argument constructors for `DbConnection::ExecuteQuery()`, so I renamed the existing constructors to static factory methods:

* `Collection::FromName()`
* `Contributor::FromProperties()`
* `EbookSource::FromTypeAndUrl()`
* `GitCommit::FromLog()`

The DB path uses the zero-argument constructors, and the filesystem path uses the static factory methods.

2. (large) I added a step to `update-ebook-database` so that it:

a. creates an `Ebook` from the filesystem
b. updates the DB
c. reads back the `Ebook` from the DB and ensures that the two `Ebook` instances match

It's extra work, and I'm bad at PHP reflection so the implementation could be better, but it was useful while developing and debugging. It gives me the confidence to move onto the read path tasks like updating `ebook.php` knowing that we can pull valid `Ebook` instances from the DB. My local DB has 961 `Ebook` records, and the script didn't find any object differences.

### DIFF
--- a/config/sql/se/EbookLocSubjects.sql
+++ b/config/sql/se/EbookLocSubjects.sql
@@ -1,5 +1,7 @@
 CREATE TABLE `EbookLocSubjects` (
+  `EbookLocSubjectId` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `EbookId` int(10) unsigned NOT NULL,
   `LocSubjectId` int(10) unsigned NOT NULL,
+  PRIMARY KEY (`EbookLocSubjectId`),
   UNIQUE KEY `idxUnique` (`EbookId`,`LocSubjectId`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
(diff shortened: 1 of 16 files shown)


colagrosso:
Thanks for the quick look! I'll fix the type errors. PHPStan returns no errors, but it's a good reminder to be extra careful.

PR #369, opened 2024-05-29
Various accessibility enhancements

robinwhittleton:
I ran as many templates as I could find through axe devTools, and fixed the majority of the issues in this set of commits that should make sense individually.

There were two things that seem to be either false positives or arguable. One is for scrolling code containers (e.g. terminal snippets) that apparently aren’t keyboard accessible, but I had no problems using them. The other is for various landmarks needing specific labels, but I’m not convinced at this point. Can always add it later if necessary.


acabal:
Great work, thanks!

PR #370, opened 2024-05-29
Correct errors found in #336

colagrosso:


lib/Ebook.php
```diff
@@ -529,29 +521,26 @@ protected function GetTextSinglePageUrl(): string{
 		return $this->_TextSinglePageUrl;
 	}
 
-	protected function GetTextSinglePageSizeNumber(): string{
```
colagrosso, line 532: This diff looks messy in unified mode, but easier to read in split mode (gear menu > split) because I replaced two functions with one. I think this is a nice improvement and a logical step after we decided to store the byte count in the DB.

acabal:
Looks good, thanks!

PR #371, opened 2024-05-29
Read path: Use Ebook::GetByIdentifier() to read ebooks in ebook.php

colagrosso:
I'm kind of amazed that this works, but it does. Still lots to do, but `ebook.php` was easier than expected.


acabal:
Great!

PR #373, opened 2024-06-01
Fix <b> tag in things-to-look-out-for-when-proofreading.php

bob-reus:
Fix `XML Parsing Error: mismatched tag. Expected: </br>.` error, by replacing `<br>` with `<b>` tag


acabal:
Thanks!

PR #374, opened 2024-06-13
DB read path: Convert a few more straightforward cases to the DB

colagrosso:
* Downloads via `www/ebooks/download.php`
* Artworks
* List of collections, i.e., `/collections`
* List of ebooks in  a collection, e.g., `/collections/the-bbcs-100-greatest-british-novels-2015`
* List of tags for use on `templates/SearchForm.php`, which is part of `www/ebooks/index.php`

### DIFF
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -472,7 +472,7 @@ class Ebook{
 				$this->_ContributorsHtml .= ' illustrated by ' . Ebook::GenerateContributorList($this->Illustrators, false) . ';';
 			}
 
-			if($this->_ContributorsHtml !== null){
+			if(!empty($this->_ContributorsHtml)){
 				$this->_ContributorsHtml = ucfirst(rtrim(trim($this->_ContributorsHtml), ';'));
 
 				if(substr(strip_tags($this->_ContributorsHtml), -1) != '.'){
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -16,6 +16,7 @@ use function Safe\shell_exec;
  * @property array<EbookTag> $Tags
  * @property array<LocSubject> $LocSubjects
  * @property array<Collection> $Collections
+ * @property array<string, int> $CollectionPositions
  * @property array<EbookSource> $Sources
  * @property array<Contributor> $Authors
  * @property array<Contributor> $Illustrators
@@ -78,6 +79,8 @@ class Ebook{
 	protected $_LocSubjects = null;
 	/** @var array<Collection> $_Collections */
 	protected $_Collections = null;
+	/** @var array<string, int> $_CollectionPositions */
+	protected $_CollectionPositions = null;
 	/** @var array<EbookSource> $_Sources */
 	protected $_Sources = null;
 	/** @var array<Contributor> $_Authors */
@@ -172,13 +175,36 @@ class Ebook{
 	protected function GetCollections(): array{
 		if($this->_Collections === null){
 			$this->_Collections = Db::Query('
-							SELECT *
-							from Collections
-							where EbookId = ?
+							SELECT c.*
+							from Collections c
+							inner join CollectionEbooks ce using (CollectionId)
+							where ce.EbookId = ?
 						', [$this->EbookId], Collection::class);
 		}
 
 		return $this->_Collections;
+	}
+
+	/**
+	 * @return array<string, int>
+	 */
+	protected function GetCollectionPositions(): array{
+		if($this->_CollectionPositions === null){
+			$this->_CollectionPositions = [];
+
+			$result = Db::Query('
+					SELECT c.UrlName, ce.SequenceNumber
+					from Collections c
+					inner join CollectionEbooks ce using (CollectionId)
+					where ce.EbookId = ?
+				', [$this->EbookId], stdClass::class);
+
+			foreach($result as $row){
+				$this->_CollectionPositions[(string)$row->UrlName] = $row->SequenceNumber;
+			}
+		}
+
+		return $this->_CollectionPositions;
 	}
 
 	/**
@@ -750,12 +776,13 @@ class Ebook{
 
 		// Get SE collections
 		$collections = [];
+		$collectionPositions = [];
 		foreach($xml->xpath('/package/metadata/meta[@property="belongs-to-collection"]') ?: [] as $collection){
 			$c = Collection::FromName($collection);
 			$id = $collection->attributes()->id ?? '';
 
 			foreach($xml->xpath('/package/metadata/meta[@refines="#' . $id . '"][@property="group-position"]') ?: [] as $s){
-				$c->SequenceNumber = (int)$s;
+				$collectionPositions[$c->UrlName] = (int)$s;
 			}
 			foreach($xml->xpath('/package/metadata/meta[@refines="#' . $id . '"][@property="collection-type"]') ?: [] as $s){
 				$c->Type = (string)$s;
@@ -763,6 +790,7 @@ class Ebook{
 			$collections[] = $c;
 		}
 		$ebookFromFilesystem->Collections = $collections;
+		$ebookFromFilesystem->CollectionPositions = $collectionPositions;
 
 		// Get LoC tags
 		$locSubjects = [];
@@ -1125,14 +1153,19 @@ class Ebook{
 		$this->LocSubjects = $subjects;
 	}
 
-	public function GetCollectionPosition(Collection $collection): ?int{
-		foreach($this->Collections as $c){
-			if($c->Name == $collection->Name){
-				return $c->SequenceNumber;
-			}
+	/**
+	 * @throws Exceptions\ValidationException
+	 */
+	private function InsertCollections(): void{
+		$collections = [];
+		foreach($this->Collections as $collection){
+			$collections[] = $collection->GetByUrlNameOrCreate($collection->UrlName);
 		}
+		$this->Collections = $collections;
+	}
 
-		return null;
+	public function GetCollectionPosition(Collection $collection): ?int{
+		return $this->CollectionPositions[$collection->UrlName] ?? null;
 	}
 
 	public function Contains(string $query): bool{
@@ -1456,6 +1489,7 @@ class Ebook{
 
 		$this->InsertTagStrings();
 		$this->InsertLocSubjectStrings();
+		$this->InsertCollections();
 
 		Db::Query('
 			INSERT into Ebooks (Identifier, WwwFilesystemPath, RepoFilesystemPath, KindleCoverUrl, EpubUrl,
@@ -1496,10 +1530,13 @@ class Ebook{
 		$this->InsertTags();
 		$this->InsertLocSubjects();
 		$this->InsertGitCommits();
-		$this->InsertCollections();
 		$this->InsertSources();
 		$this->InsertContributors();
 		$this->InsertTocEntries();
+
+		foreach($this->Collections as $collection){
+			$collection->AddEbook($this, $this->GetCollectionPosition($collection));
+		}
 	}
 
 	/**
@@ -1510,6 +1547,7 @@ class Ebook{
 
 		$this->InsertTagStrings();
 		$this->InsertLocSubjectStrings();
+		$this->InsertCollections();
 
 		Db::Query('
 			UPDATE Ebooks
@@ -1555,9 +1593,6 @@ class Ebook{
 		$this->DeleteGitCommits();
 		$this->InsertGitCommits();
 
-		$this->DeleteCollections();
-		$this->InsertCollections();
-
 		$this->DeleteSources();
 		$this->InsertSources();
 
@@ -1566,6 +1601,11 @@ class Ebook{
 
 		$this->DeleteTocEntries();
 		$this->InsertTocEntries();
+
+		Collection::RemoveEbook($this);
+		foreach($this->Collections as $collection){
+			$collection->AddEbook($this, $this->GetCollectionPosition($collection));
+		}
 	}
 
 	private function DeleteTags(): void{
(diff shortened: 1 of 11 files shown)


lib/Ebook.php
```diff
@@ -472,7 +472,7 @@ protected function GetContributorsHtml(): string{
 				$this->_ContributorsHtml .= ' illustrated by ' . Ebook::GenerateContributorList($this->Illustrators, false) . ';';
 			}
 
-			if($this->_ContributorsHtml !== null){
+			if(!empty($this->_ContributorsHtml)){
```
acabal, line 475: Is this correct? `empty()` checks if it's set and not false. But by convention protected variables accessed by getters and setters are null if unset. We should always check against null using `!==` and `===`.

lib/Library.php
```diff
@@ -140,31 +140,49 @@ public static function GetEbooksByTag(string $tag): array{
...
 	/**
-	 * @return array<string, Collection>
+	 * @return array<Collection>
 	 * @throws Exceptions\AppException
 	 */
 	public static function GetEbookCollections(): array{
-		/** @var array<string, Collection> */
-		return self::GetFromApcu('collections');
+		$collections = Db::Query('
+					SELECT distinct UrlName, Name
```
acabal, line 148: Is `distinct` required? It should not possible to have more than one record with the same `UrlName`.

lib/Library.php
```diff
@@ -140,31 +140,49 @@ public static function GetEbooksByTag(string $tag): array{
...
 	/**
-	 * @return array<Tag>
-	 * @throws Exceptions\AppException
+	 * @return array<EbookTag>
 	 */
 	public static function GetTags(): array{
-		/** @var array<Tag> */
-		return self::GetFromApcu('tags');
+		$tags = Db::Query('
+				SELECT distinct t.*
```
acabal, line 179: Again is `distinct` required? In what case would we have more than one of the same tag attached to an ebook?

lib/Ebook.php
colagrosso: Line 462 above sets this variable to the empty string

```php
$this->_ContributorsHtml = '';
```

then appends to it if there are contributors, translators, or illustrators, so it can never be null on line 475.

I was looking for a clear way to check if the string was empty. Which of these options would read better?

```php
if($this->_ContributorsHtml != ''){
```

```php
if(strlen($this->_ContributorsHtml > 0)){
```

(This isn't a problem on the `master` branch because it uses a temporary variable, `$titleContributors`, to help set both `$ContributorsHtml` and `$TitleWithCreditsHtml`, so the null check on `$ContributorsHtml` works.)

https://github.com/standardebooks/web/blob/c6d2d77537d7e2e86a9646ba35db562406cbc856/lib/Ebook.php#L487

lib/Library.php
colagrosso: The `Collections` table is denormalized:

```
> select * from Collections limit 20;
+--------------+---------+----------------------------------------------------+-----------------------------------------------+----------------+--------+
| CollectionId | EbookId | Name                                               | UrlName                                       | SequenceNumber | Type   |
+--------------+---------+----------------------------------------------------+-----------------------------------------------+----------------+--------+
|         2072 |       3 | Sherlock Holmes                                    | sherlock-holmes                               |              3 | series |
|         2073 |       3 | Haycraft-Queen Cornerstones                        | haycraft-queen-cornerstones                   |           NULL | set    |
|         2074 |       5 | Scarlet Pimpernel                                  | scarlet-pimpernel                             |              9 | series |
|         2075 |       9 | Poirot                                             | poirot                                        |              6 | series |
|         2076 |      11 | The Guardian’s Best 100 Novels in English (2015)   | the-guardians-best-100-novels-in-english-2015 |             28 | set    |
|         2077 |      11 | The BBC’s 100 Greatest British Novels (2015)       | the-bbcs-100-greatest-british-novels-2015     |             52 | set    |
|         2078 |      13 | Chronicles of Carlingford                          | chronicles-of-carlingford                     |              4 | series |
|         2079 |      14 | Anne of Green Gables                               | anne-of-green-gables                          |              2 | series |
|         2080 |      17 | Sherlock Holmes                                    | sherlock-holmes                               |              8 | series |
|         2081 |      17 | Haycraft-Queen Cornerstones                        | haycraft-queen-cornerstones                   |           NULL | set    |
|         2082 |      18 | Voyages Extraordinaires                            | voyages-extraordinaires                       |             42 | series |
|         2083 |      19 | Modern Library’s 100 Best Novels                   | modern-librarys-100-best-novels               |             88 | set    |
|         2084 |      19 | The Guardian’s Best 100 Novels in English (2015)   | the-guardians-best-100-novels-in-english-2015 |             35 | set    |
|         2085 |      21 | Modern Library’s 100 Best Novels                   | modern-librarys-100-best-novels               |             32 | set    |
|         2086 |      21 | The Guardian’s Best 100 Novels in English (2015)   | the-guardians-best-100-novels-in-english-2015 |             36 | set    |
|         2087 |      24 | Haycraft-Queen Cornerstones                        | haycraft-queen-cornerstones                   |           NULL | set    |
|         2088 |      30 | Modern Library’s 100 Best Nonfiction               | modern-librarys-100-best-nonfiction           |             31 | set    |
|         2089 |      35 | Modern Library’s 100 Best Novels                   | modern-librarys-100-best-novels               |             69 | set    |
|         2090 |      38 | Modern Library’s 100 Best Novels                   | modern-librarys-100-best-novels               |             47 | set    |
|         2091 |      38 | The BBC’s 100 Greatest British Novels (2015)       | the-bbcs-100-greatest-british-novels-2015     |             69 | set    |
+--------------+---------+----------------------------------------------------+-----------------------------------------------+----------------+--------+
```

I could change it if you'd prefer:

* Remove the `EbookId` column
* Remove the `SequenceNumber` column
* Make the `UrlName` column unique
* Add a join table, `EbookCollections`, to match what we did with `EbookTags` (with the addition of the `SequenceNumber` column

It would make this read clearer, but a little more work to write collection data.

lib/Library.php
colagrosso: This one is a different problem: We are sharing the `Tags` table with `EbookTags` and `ArtworkTags`. If we weren't doing that, we could just select all the `Tags` and be set. Instead, we have to join with the `EbookTags` table to get the tags that are used for ebooks. When we do that join, we get records for each (ebook, tag) pair. Here's what the query looks like without `distinct`:

```
> select t.* from Tags t inner join EbookTags et using (TagId) order by t.Name limit 100;
+-------+-----------+
| TagId | Name      |
+-------+-----------+
|     7 | Adventure |
|     7 | Adventure |
|     7 | Adventure |
|     7 | Adventure |
|     7 | Adventure |
|     7 | Adventure |
|     7 | Adventure |
|     7 | Adventure |
|     7 | Adventure |
[...]
|     7 | Adventure     |
|     7 | Adventure     |
|     7 | Adventure     |
|     7 | Adventure     |
|     7 | Adventure     |
|     7 | Adventure     |
|    11 | Autobiography |
|    11 | Autobiography |
|    11 | Autobiography |
|    11 | Autobiography |
|    11 | Autobiography |
|    11 | Autobiography |
|    11 | Autobiography |
|    11 | Autobiography |
+-------+---------------+
``` 

I considered `GROUP BY` instead of `DISTINCT`, but the explain plan was the same either way:

```
> explain select distinct t.* from Tags t inner join EbookTags et using (TagId) order by t.Name;
+------+-------------+-------+--------+---------------+-----------+---------+-------------+------+----------------------------------------------+
| id   | select_type | table | type   | possible_keys | key       | key_len | ref         | rows | Extra                                        |
+------+-------------+-------+--------+---------------+-----------+---------+-------------+------+----------------------------------------------+
|    1 | SIMPLE      | et    | index  | NULL          | idxUnique | 8       | NULL        | 1314 | Using index; Using temporary; Using filesort |
|    1 | SIMPLE      | t     | eq_ref | PRIMARY       | PRIMARY   | 4       | se.et.TagId | 1    |                                              |
+------+-------------+-------+--------+---------------+-----------+---------+-------------+------+----------------------------------------------+

> explain select t.Name from Tags t inner join EbookTags et using (TagId) group by t.Name order by t.Name;
+------+-------------+-------+--------+---------------+-----------+---------+-------------+------+----------------------------------------------+
| id   | select_type | table | type   | possible_keys | key       | key_len | ref         | rows | Extra                                        |
+------+-------------+-------+--------+---------------+-----------+---------+-------------+------+----------------------------------------------+
|    1 | SIMPLE      | et    | index  | NULL          | idxUnique | 8       | NULL        | 1314 | Using index; Using temporary; Using filesort |
|    1 | SIMPLE      | t     | eq_ref | PRIMARY       | PRIMARY   | 4       | se.et.TagId | 1    |                                              |
+------+-------------+-------+--------+---------------+-----------+---------+-------------+------+----------------------------------------------+

```

colagrosso: Oh, and sorry if this wasn't clear: The purpose of this `Library::GetTags()` is to get all the tags so that they can be shown on the search form. It's not to get the tags for a single ebook, which is `Ebook::GetTags()`:

https://github.com/standardebooks/web/blob/f33f219ae57bfcf50fe5d8f329f82fa3e36881c4/lib/Ebook.php#L138

I could rename this one from `Library::GetTags()` to `Library::GetAllEbookTags()` if that would help.

lib/Library.php
acabal: Yes this needs to be two tables:

Collections: CollectionId, Name, UrlName, Type
CollectionEbooks: EbookId, CollectionId, SequenceNumber

Then we can have:

````php
class Ebook{
    public array<Collection> $Collections;
    public function GetSequenceNumber(Collection $collection): ?int{
        // Calculate collection number 
    }
    ...
}

class Collection{
    public array<Ebook> $Ebooks
    ...
}
````

It's not *great* to need a function to get the sequence number but doing it Active Record style would mean creating a subclass which I'm also not a big fan of.

File: lib/Library.php (hunk 3 above)
Comment (acabal) [reply lib/Library.php:179]: OK. Let's add a `Type` column to the `Tags` table, which will be `enum('ebook','artwork')`. Then we can select on that column and things will be a lot easier. It's unlikely artwork and ebook tags will ever conflict anyway.

**Review by acabal — COMMENTED** [review]
See notes

**Comment by colagrosso** [comment]
Thanks for the review! 

I addressed the feedback about `Collections`, so I'll reply to the comment thread above about that.

I didn't update the `Tags` table yet.

File: lib/Library.php (hunk 2 above)
Comment (colagrosso) [reply lib/Library.php:148]: Here are my two commits to address this:

https://github.com/standardebooks/web/pull/374/commits/499485366c36feff3fbe81c965039c6ef7a21cd7: Update the DB schema
https://github.com/standardebooks/web/pull/374/commits/a4ce4a73421376de98532ab211e9533a83270ee2: Update the PHP logic

The `Ebook` class already had:

```php

class Ebook{
    public function GetCollectionPosition(Collection $collection): ?int{}
}
```

so I refactored that instead of adding the `Collection::GetSequenceNumber(int $ebookId)` function proposed above. I hope that's ok, because when I tried to do it the other way I ran into some problems when reading the ebook from the filesystem and populating the DB: The `Collection` object doesn't have sequence number property anymore, which is needed to write the ebook to the DB. I tried a few hacks, but I thought they looked worse than what's here.

I was also able to get away with not adding this array of ebooks property:

```php
class Collection{
    public array<Ebook> $Ebooks;
    ...
}
```

colagrosso:
Thanks for thinking about this more and the two options. I like what you wrote about not adding a new term and being consistent with the code, so I went with the second suggestion. I removed the term `membership` and added `Collection::AddEbook` and `Collection::RemoveEbook` methods.

761e01c5fa8f98e48180478dc7201d55e9192e95 is the commit with the code to review.

We haven't talked too much about this passage from `Ebook::FromFilesystem`, but I could use your guidance there for naming and readability:

```php
		// Get SE collections
		$collections = [];
		$collectionPositions = [];
		foreach($xml->xpath('/package/metadata/meta[@property="belongs-to-collection"]') ?: [] as $collection){
			$c = Collection::FromName($collection);
			$id = $collection->attributes()->id ?? '';

			foreach($xml->xpath('/package/metadata/meta[@refines="#' . $id . '"][@property="group-position"]') ?: [] as $s){
				$collectionPositions[$c->UrlName] = (int)$s;
			}
			foreach($xml->xpath('/package/metadata/meta[@refines="#' . $id . '"][@property="collection-type"]') ?: [] as $s){
				$c->Type = (string)$s;
			}
			$collections[] = $c;
		}
		$ebookFromFilesystem->Collections = $collections;
		$ebookFromFilesystem->CollectionPositions = $collectionPositions;
```

On `master`, we previously stashed the sequence number on the `Collection` object, but that's not an option anymore:

https://github.com/standardebooks/web/blob/486167dd431a3ce96b2e83a64e8f05777362616e/lib/Ebook.php#L279

Storing the collection positions in this array seems reasonable to me:

```php
	/** @var array<string, int> $_CollectionPositions */
	protected $_CollectionPositions = null;
```

but I could make a new object out of it. It would be similar to the `CollectionMembership` object you proposed above, potentially with a different name. Having the array keys be the `UrlName` is convenient, but I could work around it.

Also, I want to be sensitive to your review time. If this PR has too many commits, I could:

1. Stash the other commits about downloads, artworks, tags, etc., and use this commit to perfect `Collection`s
2. Drop the `Collection`s work from this PR, commit the more straightforward commits, then try again fresh

acabal:
Right, that's why I proposed `CollectionMembership`, because the sequence number has to go *somewhere*, but we don't have a logical place for it the way it currently stands. I suppose you could store it as in an array as a property of `Ebook` like you suggest, but then we've decoupled collections with their positions and we have to match the two by searching the array every time we want the sequence number. That's not great encapsulation and will invite errors later down the line, say if in a few years we have to revisit this code to in order to put sequence numbers somewhere else.

In any case, go ahead with how you think is best, we can always just see how things pan out in real life and if it becomes annoying we can always change it.

colagrosso:
Thanks for the extra feedback about encapsulation and future maintenance. I also appreciate the attitude that we can always change it later.

I force pushed two commits:

ec54e3ecd3e24cf548b8e199061cfe0947232ce6: Adds the `CollectionMembership` class
9e6b090b863c61865296227b14c2fb355dc6b923: Addresses the `Tag` comment above about adding a `Type` enum

Onward to some more interesting DB read path cases!

PR #381, opened 2024-07-09
Implement OpenSearch feed search with FilterEbooks

colagrosso:
I tested URLs like this

`/feeds/opds/all?query=fiction&per-page=10&page=4`
`/feeds/rss/all?query=fiction&per-page=10&page=4`
`/feeds/atom/all?query=fiction&per-page=10&page=4`

plus the previous `query`-only searches:

`/feeds/opds/all?query=roosevelt`

and they all look good.

I'm not sure how many clients are out there to those URLs, but it will be a change that the `query`-only searches will no longer return all the search results on one page.

### DIFF
--- a/www/feeds/atom/search.php
+++ b/www/feeds/atom/search.php
@@ -5,9 +5,11 @@ $ebooks = [];
 
 try{
 	$query = HttpInput::Str(GET, 'query') ?? '';
+	$startPage = HttpInput::Int(GET, 'page') ?? 1;
+	$count = HttpInput::Int(GET, 'per-page') ?? EBOOKS_PER_PAGE;
 
 	if($query !== ''){
-		$ebooks = Library::Search($query);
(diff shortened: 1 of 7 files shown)


www/feeds/atom/search.php
```diff
@@ -5,9 +5,11 @@
 
 try{
 	$query = HttpInput::Str(GET, 'query') ?? '';
+	$startPage = HttpInput::Str(GET, 'page') ?? 1;
```
acabal, line 8: This and count should be `HttpInput::Int()` not Str, in all of the search pages in this PR

acabal (changes requested):

www/feeds/atom/search.php
colagrosso: Good catch, fixed.

acabal:
Thanks!

PR #382, opened 2024-07-13
Replace LocTags with LocSubjects missed in #351

colagrosso:


colagrosso:
I added three other small, cleanup commits to this PR.

acabal:
Thanks! Since you touched the bulk downloads script, just note that the bulk downloads are a huge mess and I suggest we leave that till the very end. We can do the basic DB work so that builds/searches/ebook pages are all working, and release that live to shake it out, then later we can work on bulk downloads last.

colagrosso:
Understood, thanks for the guidance.

PR #383, opened 2024-07-14
Replace GetEbooksFromFilesystem with a DB implementation

colagrosso:
Even though you said to leave bulk downloads till the very end, this small change to `scripts/generate-bulk-downloads` allows us to get rid of a fair bit of `Library` code. The script is still a beast to run, but it doesn't run any slower after getting the ebooks from the DB instead of the filesystem.

I won't work on any further changes to bulk downloads until the rest of the site is running well with the commits in this branch.

### DIFF
--- a/lib/Library.php
+++ b/lib/Library.php
@@ -78,12 +78,13 @@ class Library{
 
 	/**
 	 * @return array<Ebook>
-	 * @throws Exceptions\AppException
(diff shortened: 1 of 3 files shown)


acabal:
OK! I think the bulk of the time in that script is just zipping up many very large zip files.

PR #389, opened 2024-08-05
Close <abbr> tag for N.B.

tipabu:
Between this and the `application/xhtml+xml` content-type the page is served as, Firefox (at least) refuses to render any of https://standardebooks.org/contribute/how-tos/things-to-look-out-for-when-proofreading


acabal:
Thanks!

PR #390, opened 2024-08-05
Add "Things to look out for when proofreading" to list of how-tos

tipabu:
Noticed this page was missing from https://standardebooks.org/contribute/how-tos while addressing #389.


acabal:
Thanks!

PR #393, opened 2024-09-01
Update wanted ebooks list

erinendrei:
Added some novels that look suitable for the first production list. 

`modernize-spelling` removes the hyphen from "home-maker", so I figured we'd likely do the same for the relevant title, but that could be discussed later if someone picks it up.


acabal:
Thanks!

PR #395, opened 2024-09-10
Ebook tags: Store UrlName so it can be queried

colagrosso:
The search form submits the `UrlName` for tags, which have a few differences compared to their `Name`, i.e.,

```html
	<option value="childrens">Children’s</option>
...
	<option value="science-fiction">Science Fiction</option>
```

This difference is causing DB searches for those two tags to fail with no results.

The `UrlName` is the correct string to query for when searching the DB, so this PR stores it in the `Tags` table.


acabal:
Thanks!

PR #396, opened 2024-09-13
Add stretched words to the list of manual typogrify checks

robinwhittleton:
The regex picks up a lot of words that don’t need non-breaking hyphens unfortunately, but I can’t think of anything better. Changing it to just vowels means ignoring “hm-m” or “stop-p-p”.


acabal:
Thanks!

PR #398, opened 2024-09-21
Update wanted ebooks list

brendanny:
remove Women in Love by D. H. Lawrence from wanted ebooks list as it's now been published


acabal:
Thanks!

PR #401, opened 2024-10-02
Merge db-rewrite into master

colagrosso:
See issue #336 for additional discussion.


lib/EbookTag.php
```diff
@@ -17,4 +22,61 @@ protected function GetUrl(): string{
...
 	}
+
+	// *******
+	// METHODS
+	// *******
+
+	/**
+	 * @throws Exceptions\ValidationException
+	 */
+	public function Validate(): void{
```
acabal, line 33: Should validate that name exists and is not null or empty. Name should also be trimmed here.

lib/Collection.php
```diff
@@ -1,18 +1,104 @@
...
 
 	public function GetSortedName(): string{
 		return preg_replace('/^(the|and|a|)\s/ius', '', $this->Name);
 	}
+
+	/**
+	 * @throws Exceptions\ValidationException
+	 */
+	public function Validate(): void{
+		$error = new Exceptions\ValidationException();
```
acabal, line 56: Again, validate name

lib/Ebook.php
```diff
@@ -306,272 +801,363 @@ public function __construct(?string $wwwFilesystemPath = null){
...
+
+
+	// *******
+	// METHODS
+	// *******
+
+	/**
+	 * @throws Exceptions\ValidationException
+	 */
+	public function Validate(): void{
```
acabal, line 941: Here we must check that the properties you compare are `isset()` first .If they aren't set, then testing against `''` will be a fatal error. You should also trim everything here as well.

lib/Ebook.php
```diff
@@ -797,23 +1383,318 @@ private function NullIfEmpty($elements): ?string{
...
+			EbookId = ?
+		', [$this->EbookId]
+		);
+	}
+
+	private function InsertGitCommits(): void{
+		foreach($this->GitCommits as $commit){
+			$commit->EbookId = $this->EbookId;
+
+			Db::Query('
```
acabal, line 1609: This should be in `GitCommit::Create()` not in `Ebook`

lib/Ebook.php
```diff
@@ -797,23 +1383,318 @@ private function NullIfEmpty($elements): ?string{
...
+	private function InsertSources(): void{
+		foreach($this->Sources as $source){
+			$source->EbookId = $this->EbookId;
+
+			Db::Query('
+				INSERT into EbookSources (EbookId, Type, Url)
+				values (?,
+					?,
+				        ?)
+			', [$source->EbookId, $source->Type->value, $source->Url]);
```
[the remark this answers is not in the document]
acabal: This should also be `EbookSource::Create()` not part of `Ebook`

lib/Ebook.php
```diff
@@ -797,23 +1383,318 @@ private function NullIfEmpty($elements): ?string{
...
+	}
+
+	private function InsertContributors(): void{
+		$allContributors = array_merge($this->Authors, $this->Illustrators, $this->Translators, $this->Contributors);
+		foreach($allContributors as $sortOrder => $contributor){
+			$contributor->EbookId = $this->EbookId;
+			$contributor->SortOrder = $sortOrder;
+
+			Db::Query('
+				INSERT into Contributors (EbookId, Name, UrlName, SortName, WikipediaUrl, MarcRole, FullName,
```
acabal, line 1661: Again, should be `Contributor::Create()`.

acabal (commented):
Some preliminary comments to address before I do a more detailed review

lib/EbookTag.php
```diff
@@ -1,14 +1,19 @@
 <?
 class EbookTag extends Tag{
-	public function __construct(string $name){
-		$this->Name = $name;
-		$this->UrlName = Formatter::MakeUrlSafe($this->Name);
-		$this->_Url = '/subjects/' . $this->UrlName;
+	public function __construct(){
+		$this->Type = 'ebook';
```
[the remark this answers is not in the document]
colagrosso: Done.

lib/LocSubject.php
```diff
@@ -0,0 +1,52 @@
+<?
+class LocSubject{
+	public int $LocSubjectId;
+	public string $Name;
+
+	/**
+	 * @throws Exceptions\ValidationException
+	 */
+	public function Validate(): void{
```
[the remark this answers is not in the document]
colagrosso: After adding `trim($this->Name)`, PHPStan throws an error when calling `isset($this->Name)`:

```
 ------ ----------------------------------------------------------------- 
  Line   lib/LocSubject.php                                               
 ------ ----------------------------------------------------------------- 
  14     Property LocSubject::$Name (string) in isset() is not nullable.  
 ------ ----------------------------------------------------------------- 
```

I believe this is due to a change in PHP 8, which also adds a deprecation warning if calling `trim()` on `null`. It looks like a good way to handle the name and avoid the deprecation warning is:

```php
$this->Name = trim($this->Name ?? '');
if($this->Name == ''){
    [...]
}
```

so I did that.

lib/Collection.php
```diff
@@ -1,18 +1,104 @@
...
+
+		if(strlen($this->Name) > EBOOKS_MAX_STRING_LENGTH){
+			$error->Add(new Exceptions\StringTooLongException('Collection name: '. $this->Name));
+		}
+
+		if($this->Type !== null && ($this->Type != CollectionType::Series && $this->Type != CollectionType::Set)){
+			$error->Add(new Exceptions\InvalidCollectionTypeException($this->Type->value));
```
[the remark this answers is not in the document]
colagrosso: Done.

lib/Ebook.php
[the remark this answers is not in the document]
colagrosso: > Pass the enum here, not `->value`

I see you added support for enums in `Db::Query()` a long time ago. Thanks!

> This should also be `EbookSource::Create()` not part of `Ebook`

Done.

lib/EbookTag.php
colagrosso: Done, with the same comment about 

```php
$this->Name = trim($this->Name ?? '');
```

above.

lib/Collection.php
colagrosso: Done.

lib/Ebook.php
colagrosso: Could you clarify something about these properties? It seems there are some that are truly required and marked `NOT NULL` in the DB schema:

* `Identifier`
* `WwwFilesystemPath`
* `RepoFilesystemPath`
* `Title`
* `Description`
* `LongDescription`
* `WordCount`
* `ReadingEase`
* `EbookCreated`
* `EbookUpdated`
* `TextSinglePageByteCount`
* `IndexableText`

For those properties, I trimmed the string ones and checked the others with `isset()`.

Then there are properties that don't seem strictly required and are allowed to be `NULL` in the DB schema:

* `KindleCoverUrl`
* `EpubUrl`
* `AdvancedEpubUrl`
* `KepubUrl`
* `Azw3Url`
* `DistCoverUrl`
* `FullTitle`
* `AlternateTitle`
* `GitHubUrl`
* `WikipediaUrl`

For those, I trimmed them if they are not `null`, but I didn't coerce them from `null` to `''`. That seems right to me.  Some of those URLs don't seem optional to me, though, but maybe formats change over time.

One place where I changed my mind was the `Language` property. It's currently allowed to be `NULL` in the DB, but that seems wrong. I changed the code to not allow a `null` in `$this->Language`, and if you agree I'll update the DB schema, too.

In my 1000+ ebooks, all of them have `Language` set:

```
> select * from Ebooks where Language IS NULL\G
```

lib/Ebook.php

lib/Ebook.php

lib/Ebook.php
acabal: In general, if it's `NOT NULL` in the DB then the property *must* have a value during validation.

Also in general, all string properties should be trimmed during validation.

For properties that *are* nullable, if it's a string, then an empty string should be changed to `null` before saving. This example covers all cases:

````php
function Validate(){
    $this->NullableStringProperty = trim($this->NullableStringProperty ?? '');
    if($this->NullableStringProperty == ''){
        $this->NullableStringProperty = null;
    }

    if(!isset($this->NotNullableProperty)){
        // Validation error - property that is not nullable has not been set before validation.
    }
}
````

Does that answer your question?

Comment (colagrosso) [reply lib/Ebook.php:941]: > Does that answer your question?

Yep, thanks a bunch for the example code. I that approach to the `Validate()` methods in `Ebook`, `Collection`, `Contributor`, `EbookSource`, `EbookTag`, `GitCommit`, and `LocSubject`. Regarding the "Possible validation error?" for non-nullable string properties, I treated them as validation errors with the same exception as if they property was `null`. That seemed like what we wanted.

File: lib/Collection.php [hunk 10]
```diff
@@ -1,18 +1,130 @@
...
+	public int $CollectionId;
 	public string $Name;
-	public string $Url;
-	public ?int $SequenceNumber = null;
-	public ?string $Type = null;
+	public string $UrlName;
+	public ?CollectionType $Type = null;
+	protected ?string $_Url = null;
+
+	protected function GetUrl(): ?string{
```
acabal, line 16: Return value should not be nullable

lib/Constants.php
```diff
@@ -28,10 +28,16 @@
...
+const EBOOKS_IDENTIFIER_PREFIX =	'url:https://standardebooks.org/ebooks/';
+
 const DATABASE_DEFAULT_DATABASE = 	'se';
 const DATABASE_DEFAULT_HOST = 		'localhost';
 
 const EBOOKS_PER_PAGE = 12;
+const EBOOKS_MAX_STRING_LENGTH = 250;
+const EBOOKS_MAX_LONG_STRING_LENGTH = 500;
+const EBOOK_SINGLE_PAGE_SIZE_WARNING = 3 *1024 * 1024; // 3145728
+const EBOOK_EARLIEST_CREATION_DATE = new DateTimeImmutable('2014-01-01');
```
acabal, line 40: I don't think we need to have this constant or check against it, seems strange to hard code the earliest ebook date.

lib/Ebook.php
```diff
@@ -797,23 +1535,286 @@ private function NullIfEmpty($elements): ?string{
...
+			DELETE from EbookLocSubjects
+			where EbookId = ?
+		', [$this->EbookId]
+		);
+	}
+
+	private function InsertLocSubjects(): void{
+		foreach($this->LocSubjects as $locSubject){
+			Db::Query('
+				INSERT into EbookLocSubjects (EbookId, LocSubjectId)
```
acabal, line 1714: Also will throw a duplicate key exception... check all functions in which we add a relation to an ebook for this

lib/EbookSource.php
```diff
@@ -1,10 +1,61 @@
...
 class EbookSource{
+	public ?int $EbookSourceId = null;
+	public ?int $EbookId = null;
 	public EbookSourceType $Type;
 	public string $Url;
 
-	public function __construct(EbookSourceType $type, string $url){
-		$this->Type = $type;
-		$this->Url = $url;
+	public static function FromTypeAndUrl(EbookSourceType $type, string $url): EbookSource{
```
acabal, line 11: I think this function is unnecessary - if we're creating a new instance, we can just inline the contents of this function instead of having this wrapper because there is no special logic here.

lib/GitCommit.php
```diff
@@ -2,21 +2,90 @@
...
+	public ?int $EbookId = null;
 	public DateTimeImmutable $Created;
 	public string $Message;
 	public string $Hash;
 
 	/**
 	 * @throws Exceptions\InvalidGitCommitException
 	 */
-	public function __construct(string $unixTimestamp, string $hash, string $message){
+	public static function FromLog(string $unixTimestamp, string $hash, string $message): GitCommit{
```
acabal, line 14: I suggest we refactor this to accept a string, since we're saying "create this object from a log line" and not "create this object from several structured parameters". The latter is more akin to `__construct`. Then this function would parse the log line string and create the object. Or, rename this to `__construct`.

acabal (changes requested):
OK, I've had some time to do a more in depth review. Basically it looks very good. I think we're almost ready to go. I made some comments on some minor issues, mostly refactoring some names for clarity. Let me know when it's all wrapped up and then we can release it live to see how things go.

acabal:
Also, I don't think the `TocEntries` table needs an ID column. We never select on ID and it's a one to many relationship anyway. We *do* order by TocEntryId, but that's wrong - we should have a `SortOrder` column instead, and sort by that.

Likewise for `CollectionEbookId`, `EbookTagId`, `EbookSourceId`, `EbookLocSubjectId`, etc. Those columns are not necessary and if we sort by any of them, we should have a `SortOrder` column instead.

colagrosso:
Thanks for the thorough review!

> Let me know when it's all wrapped up and then we can release it live to see how things go.

I'll ping this PR again after I test it all again to make sure it's all wrapped up.

For now, I'm just responding to your comments above. My responses might require another round of review.

lib/Ebook.php
```diff
@@ -306,272 +801,515 @@ public function __construct(?string $wwwFilesystemPath = null){
...
+			$error->Add(new Exceptions\EbookRepoFilesystemPathRequiredException());
+		}
+
+		$this->KindleCoverUrl = trim($this->KindleCoverUrl ?? '');
+		if($this->KindleCoverUrl == ''){
+			$this->KindleCoverUrl = null;
+		}
+
+		if(isset($this->KindleCoverUrl)){
+			if(!preg_match('|/*_EBOK_portrait.jpg|ius', $this->KindleCoverUrl)){
```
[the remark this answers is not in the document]
colagrosso: Good catch, done.

lib/Ebook.php
```diff
@@ -306,272 +801,515 @@ public function __construct(?string $wwwFilesystemPath = null){
...
-		// Figure out ancillary links.
+		$this->EpubUrl = trim($this->EpubUrl ?? '');
+		if($this->EpubUrl == ''){
+			$this->EpubUrl = null;
+		}
 
-		// First the Wikipedia URLs.
-		$this->WikipediaUrl = $this->NullIfEmpty($xml->xpath('/package/metadata/meta[@property="se:url.encyclopedia.wikipedia"][not(@refines)]'));
+		if(isset($this->EpubUrl)){
+			if(!preg_match('|/*.epub|ius', $this->EpubUrl)){
```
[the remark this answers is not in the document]

lib/Ebook.php
```diff
@@ -306,272 +801,515 @@ public function __construct(?string $wwwFilesystemPath = null){
...
-				$this->Sources[] = new EbookSource(EbookSourceType::ProjectGutenbergAustralia, $e);
+		}
+
+		$this->AdvancedEpubUrl = trim($this->AdvancedEpubUrl ?? '');
+		if($this->AdvancedEpubUrl == ''){
+			$this->AdvancedEpubUrl = null;
+		}
+
+		if(isset($this->AdvancedEpubUrl)){
+			if(!preg_match('|/*_advanced.epub|ius', $this->AdvancedEpubUrl)){
```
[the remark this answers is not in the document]
colagrosso: Done for all uses of `preg_match`.

lib/Ebook.php
```diff
@@ -306,272 +801,515 @@ public function __construct(?string $wwwFilesystemPath = null){
...
 
-		$searchString = $this->FullTitle ?? $this->Title;
+		$this->GitHubUrl = trim($this->GitHubUrl ?? '');
+		if($this->GitHubUrl == ''){
+			$this->GitHubUrl = null;
+		}
 
-		$searchString .= ' ' . $this->AlternateTitle;
+		if(isset($this->GitHubUrl)){
+			if(!preg_match('|https://github.com/standardebooks/\w+|ius', $this->GitHubUrl)){
```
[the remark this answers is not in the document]

lib/Ebook.php
```diff
@@ -306,272 +801,515 @@ public function __construct(?string $wwwFilesystemPath = null){
...
+		}
+		catch(Exceptions\EbookNotFoundException){
+			$this->Create();
+		}
+	}
+
+	/**
+	 * @throws Exceptions\ValidationException
+	 */
+	private function InsertTagStrings(): void{
```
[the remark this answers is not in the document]
colagrosso: Yep, I renamed the functions to follow the new `Ebook::CreateTags()` and `Ebook::AddTags()` pattern.

lib/Ebook.php
```diff
@@ -797,23 +1535,286 @@ private function NullIfEmpty($elements): ?string{
...
+			DELETE from EbookTags
+			where EbookId = ?
+		', [$this->EbookId]
+		);
+	}
+
+	private function InsertTags(): void{
+		foreach($this->Tags as $tag){
+			Db::Query('
+				INSERT into EbookTags (EbookId, TagId)
```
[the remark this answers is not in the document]
colagrosso: I added the check, but I wanted to point out that it's not possible to get a  `DuplicateDatabaseKeyException` from the code as it works today. `Ebook::Save()` calls:

```php  
$this->RemoveTags();  
$this->AddTags();  
```

so any previous entries in `EbookTags` are gone when `AddTags()` runs. I can elaborate on why I like this approach if it would help. Same approach for `LocSubjects`, `CollectionMemberships`, `GitCommits`, `Sources`, `Contributors`, and `TocEntries`.

lib/CollectionMembership.php
```diff
@@ -0,0 +1,15 @@
+<?
+use function Safe\preg_replace;
```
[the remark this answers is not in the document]
colagrosso: Done.

lib/Ebook.php
```diff
@@ -48,71 +65,574 @@ class Ebook{
...
+	/**
+	 * @return array<CollectionMembership>
+	 */
+	protected function GetCollectionMemberships(): array{
+		if($this->_CollectionMemberships === null){
+			$this->_CollectionMemberships = Db::Query('
+							SELECT *
+							from CollectionEbooks
+							where EbookId = ?
+							order by CollectionEbookId
```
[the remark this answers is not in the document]
colagrosso: As you pointed out below, I removed `CollectionEbookId` and added a `SortOrder` to this table because that's what we want to sort on, not `SequenceNumber`. The `SortOrder` presents the collection memberships in the same order the producer added them to `content.opf`, e.g. for *To the Lighthouse*:

№ 15 in the Modern Library’s 100 Best Novels set.  
Part of the Encyclopædia Britannica’s Great Books of the Western World set.  
№ 2 in the BBC’s 100 Greatest British Novels (2015) set.

(The sequence numbers there are 15, `NULL`, 2, so they're not what we want to sort on here. We do sort by `SequenceNumber` in `collections/get.php` to present the ebooks in the right order in a collection.)

lib/Ebook.php
```diff
@@ -48,71 +65,574 @@ class Ebook{
...
+
+			$index = 0;
+			while($bytes >= 1024 && $index < count($sizes) - 1){
+				$bytes /= 1024;
+				$index++;
+			}
+
+			if($index == 0){
+				// No decimal point for smaller than a KB.
+				$this->_TextSinglePageSizeFormatted = sprintf("%d %s", $bytes, $sizes[$index]);
```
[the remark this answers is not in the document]
colagrosso: I removed the `Safe\sprintf`line above. In PHP8,  `sprintf()` can throw `ValueError` and `ArgumentCountError`:  
   
https://www.php.net/manual/en/function.sprintf.php\#refsect1-function.sprintf-errors

but it's not possible for this code to throw those two exceptions. I didn't add `catch` blocks here, but I will if needed.

scripts/update-ebook-database
```diff
@@ -0,0 +1,55 @@
...
+
+if($verbose){
+	print("Title: $ebook->Title\n");
+	print("Identifier: $ebook->Identifier\n");
+}
+
+try{
+	$ebook->CreateOrUpdate();
+}
+catch(Exceptions\ValidationException $validationException){
```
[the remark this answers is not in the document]
colagrosso: Running `update-ebook-database --ebookWwwFilesystemPath /does/not/exist` causes an exception in this line above:

```php  
$ebook = Ebook::FromFilesystem($ebookWwwFilesystemPath);    
```

which used to print a fatal error and exit:

```
PHP Fatal error:  Uncaught Exceptions\\EbookNotFoundException: Invalid www filesystem path: /does/not/exist in /standardebooks.org/web/lib/Ebook.php:623  
Stack trace:  
\#0 /standardebooks.org/web/scripts/update-ebook-database(26): Ebook::FromFilesystem()  
\#1 {main}  
  thrown in /standardebooks.org/web/lib/Ebook.php on line 623  
```

but now I added a `try/catch` around that line to print a custom error message:

```
An AppException occurred when updating the ebook database:  
        Exceptions\\EbookNotFoundException: Invalid www filesystem path: /does/not/exist  
```

I think that's what you're asking for. Let me know if I should add the stack trace or any other info to the error message.

The line you commented on catches only `ValidationException`. I'm not expecting any other types of exceptions from `CreateOrUpdate()`, but I could add another `catch` for `AppException` if you think it's a good idea.

scripts/update-ebook-database
```diff
@@ -0,0 +1,55 @@
...
+
+$ebookWwwFilesystemPath = $options['ebookWwwFilesystemPath'] ?? null;
+
+$verbose = false;
+if(isset($options['v']) || isset($options['verbose'])){
+	$verbose = true;
+}
+
+if($ebookWwwFilesystemPath === null){
+	print("Expected usage: update-ebook-database [-v,--verbose] --ebookWwwFilesystemPath <path>\n");
```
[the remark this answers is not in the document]
colagrosso: Done.

lib/Collection.php
colagrosso: Done.

lib/Constants.php
colagrosso: Done. I thought it would be useful to check for garbage dates that were way too early, e.g., 1970-01-01, but that hasn't been a problem so far.

lib/Ebook.php
colagrosso: Done throughout, with the same caveat above that it's not possible today to throw a `DuplicateDatabaseKeyException`.

lib/EbookSource.php
colagrosso: Done.

lib/GitCommit.php
colagrosso: Done.

colagrosso:
> Also, I don't think the `TocEntries` table needs an ID column. We never select on ID and it's a one to many relationship anyway. We _do_ order by TocEntryId, but that's wrong - we should have a `SortOrder` column instead, and sort by that.
> 
> Likewise for `CollectionEbookId`, `EbookTagId`, `EbookSourceId`, `EbookLocSubjectId`, etc. Those columns are not necessary and if we sort by any of them, we should have a `SortOrder` column instead.

I removed these 7 primary keys:

* `CollectionEbookId`  
* `ContributorId`  
* `EbookLocSubjectId`  
* `EbookSourceId`  
* `EbookTagId`  
* `GitCommitId`  
* `TocEntryId`

In almost all cases, I added a `SortOrder` column. (`Contributors` already had one. `GitCommits` already sorts on `Created`.)

I believe it's important to sort the collection memberships, tags, sources, etc. by the order the producer listed them in `content.opf`. This matches the site's current behavior.

acabal:
OK great! Let me know when you've tested again and you think this is ready to go. I'll start testing then.

colagrosso:
Ok, all set! I tested it again on a new droplet, and it worked well. Anecdotally, performance seemed similar to the production site.

There's always room for improvement with my code, but I'm happy stamping this as an initial DB milestone.

acabal:
OK, this branch is now live on the site! The site is currently serving ebooks from the DB.

I'll let it shake out for a few days. If everything looks good we can merge this in and I'll make an official announcement!

robinwhittleton:
Can confirm that a search for Poe is looking loads better. Thanks for all the hard work!

acabal:
I'm coming across an error involving artworks:

If an approved artwork is assigned an SE ebook identifier, but the ebook is not released yet, getting the artwork crashes because the ebook is not in the database.

For example `https://standardebooks.org/artworks/anonymous/down-where-the-glades-begin`.

Here's a stack trace for that URL:

````
PHP Fatal error:  Uncaught Exceptions\EbookNotFoundException: Invalid identifier: url:https://standardebooks.org/ebooks/dashiell-hammett/the-dain-curse in /standardebooks.org/web/lib/Ebook.php:1554
Stack trace:
#0 /standardebooks.org/web/lib/Artwork.php(271): Ebook::GetByIdentifier()
#1 /standardebooks.org/web/lib/Traits/Accessor.php(41): Artwork->GetEbook()
#2 /standardebooks.org/web/templates/ArtworkStatus.php(8): Artwork->__get()
#3 /standardebooks.org/web/lib/Template.php(17): include('...')
#4 /standardebooks.org/web/lib/Template.php(29): Template::Get()
#5 /standardebooks.org/web/www/artworks/get.php(97): Template::__callStatic()
#6 {main}
  thrown in /standardebooks.org/web/lib/Ebook.php on line 1554

colagrosso:
Sorry about that. I'll get a fix and test it in a minute.

colagrosso:
Ok, sorry about that. I reproduced the issue and confirmed that db3d3b7 fixes it. I pushed that commit to this branch.

acabal:
Great, looks like it's fixed. Thanks! 

Looks like there are some merge conflicts, can you take a look when you have a minute?

colagrosso:
> Looks like there are some merge conflicts, can you take a look when you have a minute?

Yep, all done.

acabal:
OK, it's been a few days and everything looks good. I've merged it in, and everything is now live! Thanks for all your hard work!

PR #407, opened 2024-10-16
Normalise capitalisation of LoI

heyainsleymae:
The initialism's capitalisation now follows the form used in the manual.


acabal:
Thanks!

PR #410, opened 2024-10-22
Losslessly Optimize PNG images

anthonyryan1:
By using [Efficient-Compresion-Tool](https://github.com/fhanau/Efficient-Compression-Tool) we can reduce image sizes by 762.25KB (5.72%) without changing the visual appearance.

-----

Also just wanted to express gratitude for your project, it's a great social good to produce high quality books available to everyone.

A lot of what I did here could be applicable to the ebooks as well. Over half of the average ebook is going to be the cover image (because of how well the text compresses relative to  images). If you believe that would be of interest to the project I can look into making a pull request for the tools repository to optimize cover images before packing.

It would need to use a more widespread tools (`optipng` and `jpegoptim` most likely) because ECT isn't widely known or packaged. But any optimization is better than none. But it would make ebooks smaller, without compromising quality.


acabal:
Great, thanks!

As far as ebooks themselves go, the cover image JPGs are committed to the repo, so it would be up to individual producers to remember to use a compression tool. Unlike PNG, compressed JPGs can vary widely in quality so I don't think we want an automatic compressor, which would risk creating a blurry JPG automatically.

robinwhittleton:
It’s worth pointing out that we do have linting in place for excessively large covers (over 1.5mb for the 1400x2100 jpg).

PR #413, opened 2024-10-25
Add some spreadsheets and improve some others

HendrikBK:
Just added some new spreadsheets and change the links to other three, so that they editable.


acabal:
Another update just created a merge conflict, can you fix that?

PR #414, opened 2024-10-26
Add some spreadsheets and change links of others

HendrikBK:
The previous PR was deleted because I ended up rebasing  my fork while trying to resolve the merge conflict (I learnt something new, at least), apologies for that. This is one similar to the last, except that it doesn't have Marion Zimmer Bradley and Katherine MacLean, since these spreadsheets are too incomplete yet. I can add them when they are more complete.

As an aside, so far the list seems to be in order of addition, perhaps change it to alphabetical order might be preferable? Just a suggestion.


acabal:
Thanks! Order doesn't really matter I don't think, people will use 'find on page' if they're looking for something particular.

PR #415, opened 2024-10-26
Add support for Mariadb to Dockerfile

robinwhittleton:
Nothing particularly complex here. A PDO error confused me for a bit until I worked out that `php-mysql` was missing, and also from the installation instructions in the README. If I’m mistaken there please let me know.

### DIFF
--- a/vms/docker/start-server.sh
+++ b/vms/docker/start-server.sh
@@ -15,12 +15,18 @@ ln -s /standardebooks.org/web/config/php/fpm/standardebooks.test.ini /etc/php/*/
 ln -s /standardebooks.org/web/config/php/fpm/standardebooks.test.ini /etc/php/*/fpm/conf.d/
 ln -s /standardebooks.org/web/config/php/fpm/standardebooks.test.conf /etc/php/*/fpm/pool.d/
 
+# Create and populate the SE database.
+service mariadb start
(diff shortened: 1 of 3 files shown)


vms/docker/start-server.sh
```diff
@@ -15,12 +15,18 @@ ln -s /standardebooks.org/web/config/php/fpm/standardebooks.test.ini /etc/php/*/
...
+mariadb < /standardebooks.org/web/config/sql/users.sql
+mariadb se < /standardebooks.org/web/config/sql/se/*.sql
+
 # Enable web server configuration
 a2ensite standardebooks.test
 
 # Restart services to load new configuration
 service apache2 restart
-service php7.4-fpm restart
+service php8.1-fpm restart
```
acabal, line 29: What I usually do here is `service "php*-fpm" restart` so that the version number doesn't matter, because it usually changes with each new Ubuntu version.

acabal (changes requested):

vms/docker/start-server.sh
robinwhittleton: Unfortunately that doesn’t seem to work in the minimal Ubuntu the container uses:
```
# service "php*-fpm" restart
php*-fpm: unrecognized service
```

acabal:
OK!

PR #420, opened 2024-11-20
Small nit as Kindle, and Kobos support epub formats

khalifaali:


acabal:
This is not correct, see https://standardebooks.org/help/how-to-use-our-ebooks#kindle-epub

PR #423, opened 2024-11-23
Placeholders initial

colagrosso:
Lots and lots of work left to do, but I thought you should take a look at the schema and the initial form to catch any early mistakes. 

I incorporated what I saw in [this discussion](https://groups.google.com/g/standardebooks/c/NkifJymeA9I) on the mailing list.

Here's a screenshot of what the form looks like so far:

Screenshot_2024-11-23_14-49-12

I don't have the post handler yet, but I'm looking forward to using `Traits\PropertyFromHttp`!


lib/Ebook.php
```diff
@@ -1153,61 +1145,41 @@ public function Validate(): void{
...
-				$error->Add(new Exceptions\StringTooLongException('Ebook Language: ' . $this->Language));
-			}
+		$this->Language = trim($this->Language ?? '');
+		if($this->Language == ''){
+			$this->Language = null;
 		}
-		else{
-			$error->Add(new Exceptions\EbookLanguageRequiredException());
+
+		if(isset($this->Language) && strlen($this->Language) > 10){
```
acabal, line 1167: Make this a nested if, because we later we might want to add more checks if `Language` is not null.

templates/EbookPlaceholderForm.php
```diff
@@ -0,0 +1,128 @@
+<?
+/**
+ * @var ?Ebook $ebook
+ */
+
+if($ebook === null){
+	$ebook = new Ebook();
+	//$ebook->EbookPlaceholder = new EbookPlaceholder();
+}
```
acabal, line 9: Pattern is `$ebook = $ebook ?? new Ebook()`.

www/placeholders/new.php
```diff
@@ -0,0 +1,67 @@
+<?
+use function Safe\session_unset;
```
acabal, line 2: Let's move this file to `/ebook-placeholders/new` for consistency

acabal (commented):
Looks very good so far! Thanks for keeping it up to date with the new user edit form. Just a few minor style suggestions. Otherwise good to continue.

acabal:
Something that has already come up is how to handle adding ebooks with more than one author or translator. Thoughts?

lib/Ebook.php

templates/EbookPlaceholderForm.php
colagrosso: Done. I might have to change this again when I get the session and error handling working like on the artworks form.

www/placeholders/new.php
colagrosso: Done.

colagrosso:
Thanks for the review! I'm going to merge now and keep working on the `POST` handler.

> Something that has already come up in the spreadsheet is how to handle adding ebooks with more than one author or translator. Thoughts?

Two thoughts:

1. It would be no trouble and not too messy to add more fields to the form for additional authors, translators, and collections. Two of each would handle almost all cases. Autocomplete will keep working if each additional artist/translator/collection is in its own field.  
2. When a book finally gets produced and reviewed, that process could change the authors, translators, and the order in which they are listed, right? Or are those pretty set in stone at the outset? I ask because if the final `<dc:identifier/>` doesn't match the placeholder's `Ebook->Identifier`, then the placeholder won't get automatically removed when the book is released. We'll need to get the placeholder edit/delete form working to handle that. In fact, on the edit placeholder form, it might be easier to allow admins to edit the `Ebook->Identifier` directly than to keep futzing with the authors and translators to get it right.

Related question: Once the spreadsheet is in good shape, are you (and others) going to fill out the form manually or are you going to `POST` to `/ebook-placeholders` programmatically?

acabal:
OK, that sounds fine. We must assume that the identifier won't change, that's the only way we know which ebook is which. So if it *does* change during production then it's OK to have to delete the placeholder manually.

I think the spreadsheet will just be converted to an SQL query. Haven't thought about it yet.

PR #425, opened 2024-11-27
Revise blog post wording

bentley:
- The original 28‐year limit came from the Copyright Act of 1790, not the Constitution
- Terms were continually extended, not continuously

I tried to improve the accuracy a little bit without making it too much more complicated.


bentley:
I’ve made thematically similar social media posts the last several PD days. But I didn’t want to change the character of the post. Some general remarks:

- I think death+70 is being portrayed too positively. As you know, even the ridiculous 95‐year terms sometimes expire before that. The pre‐2019 PD drought resulted from term extension, not the US’s long avoidance of death+ terms. Also, remember that works published since 1978 are death+70 in the US, and that’s not really an improvement.
- *Eldred v. Ashcroft* ruled that repeated extensions don’t violate the “limited Times” clause. A boneheaded decision for sure. (Thomas is the only justice remaining from that decision; the other copyright maximalists were Ginsburg, Rehnquist, O’Connor, Scalia, Kennedy, and Souter.)
- Canada is currently going through a PD drought like the US did, as in 2022 they extended copyrights from death+50 to death+70 in order to align with the US.
- Copyright laws continually ratchet worldwide through reciprocity. The Canadian extension is one example; and the US extensions were inspired by Europe (particularly Germany, which had death+80 for decades although it doesn’t anymore). There’s an angle here to attack international copyright treaties for flouting Constitutional supremacy, if you’re into that.

bentley:
Not going to resolve the merge conflicts, but feel free to take pieces from this commit if you like.

acabal:
I'll correct the law portion separately, but this is just a general overview with a US slant, and not meant to be a lengthy screed or detailed examination of the situation across nations.

PR #427, opened 2024-12-05
Complete placeholder form and add POST handler

colagrosso:
The form to add ebook placeholders is usable.  Since last time:

* Updated the form to have multiple authors, translators, and collections
* Form error handling and keeping progress in the session
* The placeholders actually populate an `EbookPlaceholder` object and get written to the DB.

Here's a screenshot of what the form looks like now:

Screenshot_2024-12-05_00-30-05

The extra authors, translators, and collections are behind `<details>` tags. Here is what it looks like expanded:

Screenshot_2024-12-05_00-31-23

When you successfully create a placeholder, you're redirected back to the blank form (like the artwork form) and there's a link to the placeholder ebook:

Screenshot_2024-12-05_00-32-45

(Clicking on that link throws errors because ebook placeholders have so many null fields. Showing placeholders is what I'll work on next. I see you did some related work for the PD 2025 books.)

#### Bulk populating the DB with SQL statements

I downloaded a CSV from Hendrik's [spreadsheet](https://docs.google.com/spreadsheets/d/1OXHJcego9CINwWSJfEAo8aTJ7jZDbhvXshYsEcMuato/edit?usp=sharing) and wrote a script to convert it into the SQL statements that will insert the placeholders in the DB. The script is basically the POST handler logic, but it reads from CSV instead of HTTP POST. You can see the SQL statments in [this file](https://gist.github.com/colagrosso/7fc986a0844157af7cd048f5f0018541#file-se-placeholders-sql), which is part of [this gist](https://gist.github.com/colagrosso/7fc986a0844157af7cd048f5f0018541) that has the script and the CSV, too. Run it like this:

```bash
 ./convert-ebook-placeholders-from-csv-to-sql > se-placeholders.sql
```

#### Updating the DB when running `deploy-ebook-to-www` works
With an empty database, I tested creating a placeholder for a known book, and then running `deploy-ebook-to-www` on the actual book. It removed the placeholder record in the DB as expected. It follows the same pattern in `Create()` and `Save()`  where we now call `RemoveEbookPlaceholder()` and `AddEbookPlaceholder()`. The published book doesn't have an `EbookPlaceholder` object, so everything works.

### DIFF
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -981,7 +981,22 @@ class Ebook{
 	}
 
 	/**
-	 * Populate the `Identifier` property based on the `Title`, `Authors`, `Translators`, and `Illustrators`. Used when creating ebook placeholders.
+	 * Joins the `Name` properites of `Contributor` objects as a URL slug, e.g.,
+	 *
+	 * 	```
+	 * 	([0] => Contributor Object ([Name] => William Wordsworth), ([1] => Contributor Object ([Name] => Samuel Coleridge)))
+	 * 	```
+	 *
+	 * returns `william-wordsworth_samuel-taylor-coleridge`.
+	 *
+	 * @param array<Contributor> $contributors
+	 */
+	protected static function GetContributorsUrlSlug(array $contributors): string{
+		return implode('_', array_map('Formatter::MakeUrlSafe', array_column($contributors, 'Name')));
+	}
+
+	/**
+	 * Populates the `Identifier` property based on the `Title`, `Authors`, `Translators`, and `Illustrators`. Used when creating ebook placeholders.
 	 *
 	 * @throws Exceptions\InvalidEbookIdentifierException
 	 */
@@ -994,17 +1009,17 @@ class Ebook{
 			throw new Exceptions\InvalidEbookIdentifierException('Title required');
 		}
 
-		$authorString = implode('_', array_map('Formatter::MakeUrlSafe', array_column($this->Authors, 'Name')));
+		$authorString = Ebook::GetContributorsUrlSlug($this->Authors);
 		$titleString = Formatter::MakeUrlSafe($this->Title);
 		$translatorString = '';
 		$illustratorString = '';
 
 		if(isset($this->Translators) || !empty($this->Translators)){
-			$translatorString = implode('_', array_map('Formatter::MakeUrlSafe', array_column($this->Translators, 'Name')));
+			$translatorString = Ebook::GetContributorsUrlSlug($this->Translators);
 		}
 
 		if(isset($this->Illustrators) || !empty($this->Illustrators)){
-			$illustratorString = implode('_', array_map('Formatter::MakeUrlSafe', array_column($this->Illustrators, 'Name')));
+			$illustratorString = Ebook::GetContributorsUrlSlug($this->Illustrators);
 		}
 
 		$this->Identifier = EBOOKS_IDENTIFIER_PREFIX . $authorString . '/' . $titleString;
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -1024,11 +1024,11 @@ class Ebook{
 
 		$this->Identifier = EBOOKS_IDENTIFIER_PREFIX . $authorString . '/' . $titleString;
 
-		if(!empty($translatorString)){
+		if($translatorString != ''){
 			$this->Identifier .= '/' . $translatorString;
 		}
 
-		if(!empty($illustratorString)){
+		if($illustratorString != ''){
 			$this->Identifier .= '/' . $illustratorString;
 		}
 	}
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -41,6 +41,7 @@ use function Safe\shell_exec;
  * @property string $TextSinglePageUrl
  * @property string $TextSinglePageSizeFormatted
  * @property string $IndexableText
+ * @property EbookPlaceholder $EbookPlaceholder
  */
 class Ebook{
 	use Traits\Accessor;
@@ -112,6 +113,7 @@ class Ebook{
 	protected string $_TextSinglePageUrl;
 	protected string $_TextSinglePageSizeFormatted;
 	protected string $_IndexableText;
+	protected ?EbookPlaceholder $_EbookPlaceholder = null;
 
 	// *******
 	// GETTERS
@@ -314,7 +316,7 @@ class Ebook{
 
 	protected function GetUrl(): string{
 		if(!isset($this->_Url)){
-			$this->_Url = str_replace(WEB_ROOT, '', $this->WwwFilesystemPath);
+			$this->_Url = str_replace(EBOOKS_IDENTIFIER_ROOT, '', $this->Identifier);
 		}
 
 		return $this->_Url;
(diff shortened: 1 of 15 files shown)


lib/Ebook.php
```diff
@@ -1651,6 +1726,9 @@ public function Save(): void{
 
 		$this->RemoveTocEntries();
 		$this->AddTocEntries();
+
+		$this->RemoveEbookPlaceholder();
```
acabal, line 1730: Let's rethink how we do this, because we're adding two DB calls for every call to `Ebook::Save()` even if no changes were made to the placeholder. Saving a full ebook is going to be much more common than saving a placeholder so this will be a big drag.

One approach could be to pass a parameter to `Ebook::Save()` indicating if we want to save the placeholder as well. Maybe `Ebook::Save(bool $updatePlaceholder = false)`. That way a regular update won't make these DB calls but we can specify in for example `PATCH /ebook-placeholders` that we actually want to update the placeholder.

templates/EbookPlaceholderForm.php
```diff
@@ -16,30 +16,65 @@
...
-		<datalist id="translator-names">
-			<? foreach(Contributor::GetAllTranslatorNames() as $translator){ ?>
-				<option value="<?= Formatter::EscapeHtml($translator->Name) ?>"><?= Formatter::EscapeHtml($translator->Name) ?></option>
-			<? } ?>
-		</datalist>
-		<input
-			type="text"
-			name="translator-name"
-			list="translator-names"
+			value="<? if(isset($ebook->Authors) && count($ebook->Authors) > 0){ ?><?= Formatter::EscapeHtml($ebook->Authors[0]->Name) ?><? } ?>"
```
acabal, line 22: Use `sizeof()` for consistency

templates/EbookPlaceholderForm.php
```diff
@@ -60,64 +97,133 @@
...
 <fieldset>
 	<legend>Corpus details</legend>
 	<label>
 		<span>On the SE wanted list?</span>
 		<input
 			type="checkbox"
-			name="ebook-placeholder-is-wanted"/>
+			name="ebook-placeholder-is-wanted"
+			<? if(isset($ebook->EbookPlaceholder) && $ebook->EbookPlaceholder->IsWanted){ ?>checked="checked"<? } ?>
+		/>
```
acabal, line 185: You can make the following fields hidden if it is not on the wanted list, that will make the form a lot cleaner. Put them in a `<fieldset>` and use this CSS:

```css
input[name="ebook-placeholder-is-wanted"] + fieldset{
    display: none;
}
input[name="ebook-placeholder-is-wanted"]:checked + fieldset{
    display: block;
}

**Review by acabal — CHANGES_REQUESTED** [review]
Looks great! Some minor comments. 

For now let's not worry too much about the CSS appearance of placeholders, I want to think about that a little. For testing purposes you can just make them a dashed outline of a cover and I'll refine it once the actual back end is tested and ready.

Let me know when you think this is ready for testing!

File: lib/Ebook.php [hunk 4]
```diff
@@ -604,6 +606,19 @@ protected function GetIndexableText(): string{
 		return $this->_IndexableText;
 	}
 
+	protected function GetEbookPlaceholder(): ?EbookPlaceholder{
+		if(!isset($this->_EbookPlaceholder)){
+			$result = Db::Query('
```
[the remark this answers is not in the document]

lib/Ebook.php
```diff
@@ -966,6 +981,44 @@ public static function FromFilesystem(?string $wwwFilesystemPath = null): Ebook{
...
+	public function FillIdentifierFromTitleAndContributors(): void{
+		if(!isset($this->Authors) || empty($this->Authors)){
+			throw new Exceptions\InvalidEbookIdentifierException('Authors required');
+		}
+
+		if(!isset($this->Title)){
+			throw new Exceptions\InvalidEbookIdentifierException('Title required');
+		}
+
+		$authorString = implode('_', array_map('Formatter::MakeUrlSafe', array_column($this->Authors, 'Name')));
```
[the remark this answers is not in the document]
colagrosso: Done. Thanks for the suggestion on what to name the function.

lib/Ebook.php
```diff
@@ -966,6 +981,44 @@ public static function FromFilesystem(?string $wwwFilesystemPath = null): Ebook{
...
+			$translatorString = implode('_', array_map('Formatter::MakeUrlSafe', array_column($this->Translators, 'Name')));
+		}
+
+		if(isset($this->Illustrators) || !empty($this->Illustrators)){
+			$illustratorString = implode('_', array_map('Formatter::MakeUrlSafe', array_column($this->Illustrators, 'Name')));
+		}
+
+		$this->Identifier = EBOOKS_IDENTIFIER_PREFIX . $authorString . '/' . $titleString;
+
+		if(!empty($translatorString)){
```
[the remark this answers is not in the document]

lib/Ebook.php
colagrosso: Good point that full ebooks will be much more common than saving a placeholder, but I'll push back that this will add two DB calls and be a big drag.

For full ebooks, there will be just one DB call:

```  
DELETE from EbookPlaceholders  
where EbookId = ?  
```  

This is similar to what we do for `TocEntries`. Only about 13% of ebooks have `TocEntries`, but every `Save()` calls:

```  
DELETE from TocEntries  
where EbookId = ?  
```  

On my underpowered machine, I can't tell the difference between calling `RemoveEbookPlaceholder()/AddEbookPlaceholder()` and not when calling `Save()` on 1000 full ebooks (via `deploy-ebook-to-www --no-build` etc.). It takes about 45 seconds either way.

I'm going to leave this as-is unless you feel strongly about it.

templates/EbookPlaceholderForm.php
colagrosso: You're a wizard, thanks! I used slightly more-specific CSS and a different structure because I was getting a lot of conflicts with some rules in `core.css`, but it works well now.

Related question: Should the `POST` handler also ignore those fields? I bet that it should, and it would be no trouble to add.

I ask because Hendrik's [spreadsheet](https://docs.google.com/spreadsheets/d/1OXHJcego9CINwWSJfEAo8aTJ7jZDbhvXshYsEcMuato/edit?usp=sharing) has transcription URLs for a lot of books not on the wanted list (and a few notes, too). If we're going to ignore those fields, I'll ask him to not add them for books not on the wanted list. Or maybe it's fine to just leave them and we can always edit the placeholders later.

colagrosso:
Thanks for the quick review! I think this is ready to merge.

> For now let's not worry too much about the CSS appearance of placeholders, I want to think about that a little. For testing purposes you can just make them a dashed outline of a cover and I'll refine it once the actual back end is tested and ready.
> 
> Let me know when you think this is ready for testing!

Understood. I'll get something basic working to show placeholders, test it a bunch, and report back.

templates/EbookPlaceholderForm.php
acabal: Yes, if `ebook-placeholder-is-wanted` is unchecked then we should not fill those values from POST.

acabal:
OK. Let me know when it's tested and a basic visualization is ready for the front end!

templates/EbookPlaceholderForm.php
colagrosso: > Yes, if `ebook-placeholder-is-wanted` is unchecked then we should not fill those values from POST.

Got it, done.

colagrosso:
> OK. Let me know when it's tested and a basic visualization is ready for the front end!

Thanks, I'm going to merge this now and send you another PR on what I have for showing placeholders on the front end.

PR #428, opened 2024-12-07
Update wanted ebooks list: Barry Goldwater

bentley:
I was surprised to see such a recent political book pop up on PGDP, but it is in fact PD.

[Wikipedia:](https://en.wikipedia.org/wiki/The_Conscience_of_a_Conservative)

> …It helped revive the American conservative movement and make Goldwater a political star, and it has influenced countless conservatives in the United States, helping to lay the foundation for the Reagan Revolution of the 1980s.

> The 123-page book covers such topics as education, labor unions and policies, civil rights, agricultural policy and farm subsidies, social welfare programs, and income taxation. The book is considered to be a significant statement of politically and economically American conservative ideas which were to gain influence during the following decades.


acabal:
Interesting, the copyright must have not been renewed.

PR #429, opened 2024-12-10
Show placeholders on the web

colagrosso:
These look pretty bad (even for me), but it's minimal styling. I trust you can make them look better. Here are screenshots of:

Search results:
Screenshot_2024-12-09_10-08-46

Collections page:
Screenshot_2024-12-09_09-09-28

Author page:
Screenshot_2024-12-09_10-09-59

Single ebook placeholder:
Screenshot_2024-12-08_16-13-46

Here are assumptions I made that I'd like you to check:

### Placeholders are ok in
* Search results
* Collections page
* Author page
* Single book
    * `www/ebooks/get.php` checks `Ebook::IsPlaceholder()` and does `require www/ebook-placeholders/get.php`
* Sitemap
* Artwork page
    * If an artwork is assigned to a placeholder, the link now works, but it still reads "(unreleased)" after the link

### Placeholders are NOT ok in
* The related books carousel at the bottom of `www/ebooks/get.php`
   * Removed placeholders from the query in `Ebook::GetAllByRelated()`
* `scripts/generate-bulk-downloads` and `scripts/generate-feeds`
    * The scripts post-filter by checking `Ebook::IsPlaceholder()` after calling `Ebook::GetAll()`
* Feeds search (e.g., `www/feeds/opds/search.php`)
    * Added a filter parameter to `Ebook::GetAllByFilter()` to filter out placeholders
* `www/ebooks/download.php`
    * This would be rare because the user would have to add `/download` to a placeholder URL, but the code checks `Ebook::IsPlaceholder()` after calling `Ebook::GetByIdentifier()`

### DIFF
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -2057,13 +2057,19 @@ class Ebook{
 	*
 	* @return array{ebooks: array<Ebook>, ebooksCount: int}
 	*/
-	public static function GetAllByFilter(string $query = null, array $tags = [], Enums\EbookSortType $sort = null, int $page = 1, int $perPage = EBOOKS_PER_PAGE): array{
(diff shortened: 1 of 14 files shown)


acabal:
Perfect thanks!

colagrosso:
Right on, thanks for the quick review. Merging now.

Issue #430, opened 2024-12-11
Placeholders notes and known issues


colagrosso:
The `placeholders` branch is in good shape and is ready for you to take over, Alex. I've tested it a few different ways, and I'm happy with it.

Three known issues:

### 1\. `Ebooks` SQL schema

In your existing database, you'll need to modify these columns in `Ebooks` table to allow nulls:

ALTER TABLE Ebooks MODIFY COLUMN \`WwwFilesystemPath\` varchar(511) NULL;  
ALTER TABLE Ebooks MODIFY COLUMN \`RepoFilesystemPath\` varchar(511) NULL;  
ALTER TABLE Ebooks MODIFY COLUMN \`Description\` text NULL;  
ALTER TABLE Ebooks MODIFY COLUMN \`LongDescription\` text NULL;  
ALTER TABLE Ebooks MODIFY COLUMN \`Language\` varchar(10) NULL;  
ALTER TABLE Ebooks MODIFY COLUMN \`WordCount\` int(10) unsigned NULL;  
ALTER TABLE Ebooks MODIFY COLUMN \`ReadingEase\` float NULL;  
ALTER TABLE Ebooks MODIFY COLUMN \`EbookCreated\` datetime NULL;  
ALTER TABLE Ebooks MODIFY COLUMN \`EbookUpdated\` datetime NULL;  
ALTER TABLE Ebooks MODIFY COLUMN \`TextSinglePageByteCount\` bigint unsigned NULL;

### 2\. PHPStan errors to fix later

Because those columns above can now be null, PHPStan is throwing errors like:

```
 ------ ----------------------------------------------------------------------------   
  Line   lib/AtomFeed.php                                                                                                                                       
 ------ ----------------------------------------------------------------------------   
  58     Cannot call method format() on Safe\\DateTimeImmutable|null.    
 ------ ----------------------------------------------------------------------------   

 ------ ----------------------------------------------------------------------------   
  Line   scripts/generate-bulk-downloads                                      
 ------ ----------------------------------------------------------------------------   
  113    Cannot call method format() on Safe\\DateTimeImmutable|null.          
  114    Cannot call method getTimestamp() on Safe\\DateTimeImmutable|null.    
 ------ ----------------------------------------------------------------------------   

 ------ ----------------------------------------------------------------------------   
  Line   templates/EbookGrid.php                                                        
 ------ ----------------------------------------------------------------------------   
  49     Parameter \#1 $num of function number\_format expects float, int|null given.    
[...]

PR #431, opened 2024-12-12
Hide placeholders when browsing /ebooks without a search query

colagrosso:
As mentioned [here](https://github.com/standardebooks/web/issues/430#issuecomment-2536869091) in #430:

> Another thing I noticed is that placeholders appear in `/ebooks` when no search parameters are set. I think we should not show placeholders in that case - only if the user has actually entered some parameters in the search form.

This PR hides the placeholders when there is no search query. That's the only search parameter worth considering. Placeholders don't have tags, and changing the Sort, View, or Per page parameters shouldn't change whether placeholders appear. 

Two other minor commits in this PR.

### DIFF
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -1338,6 +1338,7 @@ class Ebook{
 
 	/**
 	 * @throws Exceptions\ValidationException
+	 * @throws Exceptions\DuplicateEbookException
 	 */
 	public function CreateOrUpdate(): void{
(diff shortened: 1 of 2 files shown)


acabal:
Great, thanks!

acabal:
I've had a flu all week so work has been slow. I hope to do some more testing tomorrow and then if all goes well we can release this maybe Monday.

PR #435, opened 2024-12-13
Style placeholders for list view

colagrosso:
Before
Screenshot_2024-12-12_17-49-45

After
Screenshot_2024-12-12_17-50-35

Still room for improvement in `templates/EbookGrid.php`, but this is better for list view.


acabal:
Thanks!

PR #438, opened 2024-12-15
Make placeholder outlines responsive

colagrosso:
Here are screenshots of a few different widths:

Screenshot_2024-12-15_15-26-39
Screenshot_2024-12-15_15-27-21
Screenshot_2024-12-15_15-28-33


acabal:
Great, thanks! Interesting use of `aspect-ratio`, that one's new to me.

PR #439, opened 2024-12-15
Allow author paths that don't point to www files

colagrosso:
Authors with placeholders only, e.g. (at the time of writing):

https://standardebooks.org/ebooks/dashiell-hammet

were returning 404s because there was no directory at:

/standardebooks.org/web/www/ebooks/dashiell-hammet

Now we allow calls to `Ebook::GetAllByAuthor()` for such URLs because it might find placeholders for the author.


acabal:
Thanks!

PR #441, opened 2024-12-16
Add a dropdown to specify collection type

colagrosso:
Here's a screenshot of what it looks like (showing an error case just to show that the collection type is retained):

Screenshot_2024-12-15_23-03-49


acabal:
Great! Note that we should still allow a blank type, which should set the field to null. The site will show as "collection".

I've been working on a project management system and so have been updating the database and doing some rapid force pushes. I hope I haven't screwed up your base!

acabal:
Excellent, thanks!~

PR #442, opened 2024-12-16
Nudge the placeholder ribbon left 2px, past the border

colagrosso:
To my eye, the ribbon should fold against the outside edge of the border.

Before:
Screenshot_2024-12-16_00-16-20

After:
Screenshot_2024-12-16_00-20-05

(The "In Use" ribbon on artwork looks right because it's placed on a higher element, before the border is applied.)

### DIFF
--- a/www/css/core.css
+++ b/www/css/core.css
@@ -3278,6 +3278,15 @@ ol.ebooks-list > li[property="schema:hasPart"][value].ribbon a::after{
 	top: calc(.25rem + 3px + 2rem + 4px);
 }
 
+ol.ebooks-list > li.ribbon .placeholder-cover::before, /* Ribbon */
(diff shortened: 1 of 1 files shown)


acabal:
Good catch. But let's tweak the CSS a little - we want the `.ribbon` to be reusable and the exception case is the placeholder, which has a thicker border. So we should only adjust the position for placeholders, not every single ribbon. (For example, in case in the future we want to add a ribbon to a regular ebook?)

colagrosso:
Thanks for explaining that, and I see what you mean. I changed the CSS to apply only to placeholders.

(For fun, I tried to add ribbons to regular ebooks, but they are still missing some of the styling that placeholders have. So in the future there would still be more work for regular ebooks, but these placeholder rules won't get in the way.)

acabal:
Thanks!

PR #443, opened 2024-12-16
Update wanted-ebooks.php: remove Gray, Poetry (done); update Parrish,…

dajare:
… Perennial Bachelor with transcription source, move to intermediate"


acabal:
Thanks!

PR #444, opened 2024-12-17
Remove name.vessel.airplane from se vocab

neilpa:
Duplicate with name.vehicle.airplane, and only the vehicle is allowed via `se lint`

Follow up from [mailing list](https://groups.google.com/g/standardebooks/c/ICyaNlaPj5A/m/kKa7J5_hAgAJ)


acabal:
Thanks!

PR #445, opened 2024-12-18
Add /edit endpoint to update placeholders

colagrosso:
I tested this a few ways, but there is a lot going on with `Ebook`, `EbookPlaceholder`, and `Project` objects. The `PUT` section in `www/ebook-placeholders/post.php` could benefit from an extra look to make sure it's doing what you want with `Project` objects.

One simple case that works well and you can use on the production site is to correct this placeholder:

https://standardebooks.org/ebooks/dashiell-hammet/the-maltese-falcon

to "Hammett" with two "t"s. It was wrong in the spreadsheet, but Hendrik must have corrected it recently. 

Screenshot_2024-12-17_21-10-49


acabal:
Excellent work, thanks!

I should have mentioned that we don't have to edit projects using this method. Adding projects via the placeholder form is a just convenience, because often when we add a placeholder we also want to add a project at the same time. But the two are separate concepts and will be edited separately.

I'll pull that out into its own function a little later.

colagrosso:
Thanks for explaining that, and for refactoring the editing of projects. It looks like a cool feature.

PR #447, opened 2024-12-19
Add form to delete placeholders

colagrosso:
I tested it, and it works for me, but it is pretty aggressively deleting data so be careful with it. 

I'm not sure how to style this form, so I left it pretty minimal:
image

The delete form is hidden behind the `<details>` element so that it requires two clicks:
image

### DIFF
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -2156,6 +2156,27 @@ final class Ebook{
 		}
 	}
 
+	public function Delete(): void{
+		$this->RemoveTags();
+		$this->RemoveLocSubjects();
+		$this->RemoveCollectionMemberships();
+		$this->RemoveGitCommits();
+		$this->RemoveSources();
+		$this->RemoveContributors();
+		$this->RemoveTocEntries();
+		$this->RemoveEbookPlaceholder();
+
+		foreach($this->Projects as $project){
+			$project->Delete();
+		}
+
+		Db::Query('
+			DELETE
+			from Ebooks
+			where EbookId = ?
+		', [$this->EbookId]);
+	}
+
 	// ***********
 	// ORM METHODS
 	// ***********
--- a/lib/Project.php
+++ b/lib/Project.php
@@ -455,6 +455,20 @@ final class Project{
 		}
 	}
 
+	public function Delete(): void{
+		Db::Query('
+			DELETE
+			from ProjectReminders
+			where ProjectId = ?
+		', [$this->ProjectId]);
+
+		Db::Query('
+			DELETE
+			from Projects
+			where ProjectId = ?
+		', [$this->ProjectId]);
+	}
+
 	public function FillFromHttpPost(): void{
 		$this->PropertyFromHttp('EbookId');
 		$this->PropertyFromHttp('ProducerName');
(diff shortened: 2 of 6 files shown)


colagrosso:
I added two other commits with this PR because I was getting a crash on `Ebook::GetProjects()`

```
NOTICE: PHP message: PHP Fatal error:  Uncaught Exceptions\DatabaseQueryException: Error when executing query: SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'Created' in order clause is ambiguous. Query: 
                                                     SELECT *
                                                     from Projects
                                                     inner join Ebooks
                                                     using(EbookId)
                                                     where EbookId = ?
                                                     order by Created desc
```

I believe the query worked until you added the `order by Created desc`, but even when it works it logs warnings like:

```
NOTICE: PHP message: PHP Deprecated:  Creation of dynamic property Project::$Identifier is deprecated in /standardebooks.org/web/lib/Traits/Accessor.php on line 81
NOTICE: PHP message: PHP Deprecated:  Creation of dynamic property Project::$WwwFilesystemPath is deprecated in /standardebooks.org/web/lib/Traits/Accessor.php on line 81
etc.
```

because all the fields are returned together in one big row and associated with `Project`.

But now I see that you are intending to do a multi-table select, and my commits are probably breaking your multi-table select. I don't quite understand how that works, but I'd like to. It would come in handy now it would be convenient to get `Ebook` and `EbookPlaceholder` objects from a multi-table select for the wanted list.

colagrosso:
Sorry for the spam. I looked closer into how multi-table select is supposed to work. In the case of `Projects` and `Ebooks`, I think it's not working. For example, in `www/projects/index.php`, there's a call to:

```php
$inProgressProjects = Project::GetAllByStatus(Enums\ProjectStatusType::InProgress); 
```
that runs

```sql
SELECT * from Projects inner join Ebooks using (EbookId) where Projects.Status = 'in_progress' order by Title asc
```
but then the Accessor for `Ebook::Get()` is still kicking in and running

```
SELECT * from Ebooks where EbookId = ?
```

for each `Ebook`. So the multi-table select didn't populate the `Ebook` objects like it looks like it should.

colagrosso:
Sorry again for the spam. I figured out how to get multi-table select working, as shown in this hacked up commit:

https://github.com/colagrosso/se-web/commit/9774b3a02a1582f5f4deba414527dc0760999eb7

There are three small changes:

1. Add a pass-through function, `Db::MultiTableSelect()`
2. In `Project::GetAllByStatus()`, call `Db::MultiTableSelect()` instead of `Db::Query()`
3. In `Project::FromMultiTableRow()`, check a different `EbookId` for `null`

After those changes, I confirmed that the multi-table select executed, and there were no follow-up `Ebook` queries. Pretty cool!

I dropped the commits where I modified `Ebook::Projects()`, `Ebook::GetProjectInProgress()`, and `Ebook::GetPastProjects()`.  They will still need to be converted to `Db::MultiTableSelect()`, but not as part of this PR.

acabal:
Great, yes, you definitely solved the problem. I copied and pasted that code in from a different project but forgot to include the `MultiTableSelect` bits. Entirely my fault. Thanks!

acabal:
One thing that comes up is that if we delete the last ebook in a collection, the collection remains, but is empty. We should also delete the collection in that case.

acabal:
I've fixed the multi select issues so just pull from master for those fixes.

colagrosso:
Thanks for all the follow up fixes to the delete form!

> One thing that comes up is that if we delete the last ebook in a collection, the collection remains, but is empty. We should also delete the collection in that case.

Yep, agreed. I'll get on that.

We did something similar in `Artwork.php` to delete artists that were no longer associated with any artworks:

https://github.com/standardebooks/web/blob/b4003b30e2d2ec8345d12848a48a2af14785e16e/lib/Artwork.php#L822-L829

PR #450, opened 2024-12-27
Minor tweaks

weijia-cheng:
1. Added the Georgia Douglas Johnson spreadsheet to the spreadsheets page.
2. Added a step to the review guide to review the initial commit. This is something I have been doing for some time now and I have noticed several cases of people committing the PG license into the repo. This seems like a good step to codify.


acabal:
Thanks!

PR #451, opened 2024-12-28
Add instructions to poetry how-to on manually transcribing poetry

weijia-cheng:
There are several "transcription required" advanced poetry productions on the wanted list so we may as well have some instructions for how to do it. This is what I have been using on the various omnibuses that I have worked on that require transcription and in my experience it works really well once you know how to eyeball indentation levels. Because it is so generic you can use it on anything from terrible HTML transcriptions to raw OCRed text if you need to.


EmmaSweeney:
This looks great! The only thing I would tweak is to separate this section from "Splitting Files". The title would be `<h2>` and the ToC would look like:

1. Splitting files
	1. A single collections file
	2. Multiple files
2. Manual transcription
...

weijia-cheng:
Ok, that's done now. I wasn't sure if it merited a new section.

acabal:
Thanks! Also, the `<picture>` elements should actually be `<figure class="full-width">`. Can you fix that?

PR #455, opened 2024-12-31
Make `DeleteUnused()` faster for `Collection`, `Tag`, and `LocSubject`

colagrosso:
I think the new query is readable, so I applied it to all three types.

You'll need to add these three indices to your DB:

```sql
ALTER TABLE `CollectionEbooks` ADD INDEX `index1` (`CollectionId`);
ALTER TABLE `EbookLocSubjects` ADD INDEX `index1` (`LocSubjectId`);
ALTER TABLE `EbookTags` ADD INDEX `index1` (`TagId`);
```

It runs much faster on my underpowered machine. When running just `Ebook::Save()` via:

```
$ time nice /standardebooks.org/web/scripts/deploy-ebook-to-www --verbose --no-build --no-images --no-recompose --no-epubcheck --no-feeds --no-bulk-downloads virginia-woolf_to-the-lighthouse.git
```

here are the before and after times:

Before:
```
real    0m1.682s
user    0m0.086s
sys     0m0.071s
```

After:
```
real    0m0.267s
user    0m0.085s
sys     0m0.067s
```

Here are the explain plans before and after adding the index to `EbookLocSubjects`:

```
explain delete ls from LocSubjects ls LEFT JOIN EbookLocSubjects els ON ls.LocSubjectId = els.LocSubjectId WHERE els.LocSubjectId IS NULL;
```

Before:
```
+------+-------------+-------+-------+---------------+-----------+---------+------+------+--------------------------------------+
| id   | select_type | table | type  | possible_keys | key       | key_len | ref  | rows | Extra                                |
+------+-------------+-------+-------+---------------+-----------+---------+------+------+--------------------------------------+
|    1 | SIMPLE      | ls    | ALL   | NULL          | NULL      | NULL    | NULL | 1937 |                                      |
|    1 | SIMPLE      | els   | index | NULL          | idxUnique | 8       | NULL | 8228 | Using where; Using index; Not exists |
+------+-------------+-------+-------+---------------+-----------+---------+------+------+--------------------------------------+

```

After:
```
+------+-------------+-------+------+---------------+--------+---------+--------------------+------+--------------------------------------+
| id   | select_type | table | type | possible_keys | key    | key_len | ref                | rows | Extra                                |
+------+-------------+-------+------+---------------+--------+---------+--------------------+------+--------------------------------------+
|    1 | SIMPLE      | ls    | ALL  | NULL          | NULL   | NULL    | NULL               | 1937 |                                      |
|    1 | SIMPLE      | els   | ref  | index1        | index1 | 4       | se.ls.LocSubjectId | 2    | Using where; Using index; Not exists |
+------+-------------+-------+------+---------------+--------+---------+--------------------+------+--------------------------------------+
```

### DIFF
--- a/config/sql/se/CollectionEbooks.sql
+++ b/config/sql/se/CollectionEbooks.sql
@@ -3,5 +3,6 @@ CREATE TABLE IF NOT EXISTS `CollectionEbooks` (
   `CollectionId` int(10) unsigned NOT NULL,
   `SequenceNumber` int(10) unsigned NULL,
   `SortOrder` tinyint(3) unsigned NOT NULL,
-  UNIQUE KEY `idxUnique` (`EbookId`,`CollectionId`)
+  UNIQUE KEY `idxUnique` (`EbookId`,`CollectionId`),
+  KEY `index1` (`CollectionId`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
--- a/config/sql/se/EbookLocSubjects.sql
+++ b/config/sql/se/EbookLocSubjects.sql
@@ -2,5 +2,6 @@ CREATE TABLE IF NOT EXISTS `EbookLocSubjects` (
   `EbookId` int(10) unsigned NOT NULL,
   `LocSubjectId` int(10) unsigned NOT NULL,
   `SortOrder` tinyint(3) unsigned NOT NULL,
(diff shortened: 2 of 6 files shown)


acabal:
Perfect, great work, thanks! Happy new year!

PR #457, opened 2025-01-02
Remove `templates/EbookProjects.php`

colagrosso:
378555d removed the reference to it in `www/ebook-placeholders/get.php`


acabal:
Thanks!

PR #458, opened 2025-01-02
Rewrite `MultiTableSelect` queries with `ON`

colagrosso:
Instead of `USING(EbookId)`, it would be easier to handle `MultiTableSelect` queries in `FromMultiTableRow()` if the queries used:  

```      
ON Projects.EbookId = Ebooks.EbookId
```
   
This is because `USING` will return only one `EbookId` column, but `ON` will return all columns from both tables. See [this passage of the manual](https://dev.mysql.com/doc/refman/8.0/en/join.html#:~:text=The%20USING%20join%20selects%20the%20coalesced%20value%20of%20corresponding%20columns%2C%20whereas%20the%20ON%20join%20selects%20all%20columns%20from%20all%20tables.) for details.

This removes the need to do:

```
$row['Ebooks']->EbookId = $object->EbookId;
```

in `FromMultiTableRow()` functions, which can be error prone.

No matter if the queries are written with `USING` or `ON`, it's worth adding an index to make the join more efficient, so this PR adds that, too. Here's how to add it to existing DBs:

```
ALTER TABLE `Projects` ADD INDEX `index1` (`EbookId`);
```

### DIFF
--- a/config/sql/se/Projects.sql
+++ b/config/sql/se/Projects.sql
@@ -15,5 +15,6 @@ CREATE TABLE IF NOT EXISTS `Projects` (
   `LastCommitTimestamp` DATETIME NULL DEFAULT NULL,
   `LastDiscussionTimestamp` DATETIME NULL DEFAULT NULL,
   `IsStatusAutomaticallyUpdated` tinyint(1) NOT NULL DEFAULT 1,
-  PRIMARY KEY (`ProjectId`)
+  PRIMARY KEY (`ProjectId`),
+  KEY `index1` (`EbookId`)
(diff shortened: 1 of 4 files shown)


acabal:
OK, however I think we should keep the comments in the function phpdoc, or at least clarify that it only applies to `using()`. At some point in the future we or someone else might use `using()` and be surprised at what happens, so we want to keep that documented.

colagrosso:
> OK, however I think we should keep the comments in the function phpdoc, or at least clarify that it only applies to `using()`. At some point in the future we or someone else might use `using()` and be surprised at what happens, so we want to keep that documented.

Yep, done.

acabal:
Thanks!

Issue #459, opened 2025-01-02
Contributor slugs for authors should not be generated by Validate()


acabal:
This issue just came up. We have two authors named `Samuel Butler`. One is a poet died 1680, the other is a novelist died 1902.

To distinguish the two, the older Samuel Butler's slug is `samuel-butler-1612-1680`. The younger's is `samuel-butler`.

The problem is when we create a new contributor, we create its `UrlName` in the `Contributor::Validate()` function, which makes the older's slug the same as the newer's slug. This means that when the user clicks on the older's name to see a list of his books, it returns a 404, and the younger's listing incorrectly shows the older's works.

Instead of doing that, we should extract the slug from the identifier which is presumably always correct.

(This has already been fixed in the live DB.) See https://standardebooks.org/ebooks/samuel-butler-1612-1680, https://standardebooks.org/ebooks/samuel-butler

colagrosso:
I can fix `Ebook::FromFilesystem()` to use `Identifier` to create contributors, but I won't be able to do the same for the placeholder form, i.e., `/ebook-placeholders/new`. The dropdowns in that form select contributors by `Name`, not `UrlName`, so there will be two Samuel Butlers in the dropdown but no way to specify `samuel-butler-1612-1680`. Put another way: In the placeholder form, the `Identifier` is computed *after* the contributors because it's computed based on the contributors. 

If more books will be created as placeholders first for project tracking, we could reconsider the placeholder form. I like how autocomplete works currently, though, so I'm torn.

These are just some rough thoughts, and I meant to mock up a new placeholder form with autocomplete to explore some options. Let me know if that would be helpful.

acabal:
I think that's fine, we're mostly concerned with ebooks coming into the database from actual git repos. The repos are always presumed to be totally correct. This is a very rare case so it's OK if the placeholder form isn't perfect in this regard.

PR #462, opened 2025-01-15
Match `Contributor` `UrlName` to `Identifier`

colagrosso:
I know you fixed this once in the production DB manually, but it's currently returning a 404:

https://standardebooks.org/ebooks/samuel-butler-1612-1680 

After this change, you can fix it by re-running `scripts/deploy-ebook-to-www`:

```
/standardebooks.org/web/scripts/deploy-ebook-to-www --verbose --no-build --no-images --no-recompose --no-epubcheck --no-feeds --no-bulk-downloads /standardebooks.org/ebooks/samuel-butler-1612-1680_hudibras.git 
```

`samuel-butler-1612-1680` was the only example I could find with a special suffix so far, but I added some other cases to be careful of in the docstring.

Fixes \#459

### DIFF
--- a/lib/Contributor.php
+++ b/lib/Contributor.php
@@ -83,14 +83,19 @@ class Contributor{
 				// Sometimes placeholders may have `'` in the name.
 				$this->Name = str_replace('\'', '’', $this->Name);
 			}
-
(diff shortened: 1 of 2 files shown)


acabal:
Great, thanks!

PR #463, opened 2025-01-17
Search artwork by `UrlName`

### LINKED ISSUE #461: Artwork search doesn't handle diacritics
We have an artist named `Wilhelm Heinrich Detlev Körner`. Searching for `wilhelm` returns his paintings, but searching for `körner` or `korner` returns nothing. See https://standardebooks.org/artworks/wilhelm-heinrich-detlev-korner

colagrosso:
`Artwork::GetAllByFilter()` is already removing diacritics from search terms, and `UrlName` stores the artist's name with diacritics removed by calling `Formatter::MakeUrlSafe()`.

Keeping the search by `Name` because `Formatter::MakeUrlSafe()` makes other changes to the name, too.

Fixes #461

Notes:

Searching for these terms now work:

`Körner`
`Korner`

When thinking about other edge cases and looking at  `Formatter::MakeUrlSafe()` and  `Formatter::RemoveDiacritics()`, I noticed that searching for:

`O’Keeffe`

doesn't work: https://standardebooks.org/artworks?query=O%E2%80%99Keeffe

After this PR, at least this will work:

`okeeffe`

which isn't great, but it's better.

The problem with `O’Keeffe` is that this line is turning it into `o'keeffe`

https://github.com/standardebooks/web/blob/db8d4221338669e3d47449c44733c83e42520056/lib/Artwork.php#L1032-L1034

As the comment says, it preserves apostrophes, but `Formatter::RemoveDiacritics()` is changing the apostrophe to `'` before `preg_replace()` is called. We don't have `o'keeffe` stored in the DB anywhere, just `O’Keeffe` and `okeeffe`. I don't think it's worth it do add another `replace()` to the MySQL query, so I'm going to leave it for now. We could consider an `IndexableText` field like we have for `Ebooks`.

### DIFF
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -1069,6 +1069,7 @@ class Artwork{
 			$params[] = $tokenizedQuery; // art.Name
 			$params[] = $tokenizedQuery; // art.EbookUrl
 			$params[] = $tokenizedQuery; // a.Name
+			$params[] = $tokenizedQuery; // a.UrlName
 			$params[] = $tokenizedQuery; // aan.Name
 			$params[] = $tokenizedQuery; // t.Name
 
@@ -1089,6 +1090,7 @@ class Artwork{
 				            and (art.Name regexp ?
 				            or replace(art.EbookUrl, "_", " ") regexp ?
 				            or a.Name regexp ?
+				            or a.UrlName regexp ?
 				            or aan.Name regexp ?
 				            or t.Name regexp ?)
 				    group by art.ArtworkId) x', $params);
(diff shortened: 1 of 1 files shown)


acabal:
Why did we preserve apostrophes to begin with?

If now we can search via UrlName, then does it make sense to also just remove apostrophes also, so that we `O’Keefe` matches the UrlName `okeefe`?

colagrosso:
Good point, thanks for the review! I originally preserved apostrophes in 8a8692619c5eb3024b1f5cdd58bc6aebf5b8df64 because queries like

`Whistler's Mother`

were getting converted to:

`\b(whistler|s|mother)\b`

which was a big problem because the single-character `s` matched almost everything.

Now that we know searching by `UrlName` is a good idea, I removed the apostrophes so that the query becomes:

`\b(whistlers|mother)\b`

which works. As you pointed out above, `O’Keeffe` becomes

`\b(okeeffe)\g`

which also works, so we're all set. Search is now more robust to diacritics and other non-alphanumeric characters.

acabal:
Excellent, thanks!

PR #464, opened 2025-01-17
Dashes

bentley:
Several places use hyphens when they should use dashes. This is particularly noticeable in the “Read Online” books where they show up in the page body, not just the title.

Several XSL tests rely on this, which is why I changed all instances instead of just the obvious ones. Thus, the change needs to be reviewed carefully. Unfortunately I’m unable to test a local deploy at present.

### DIFF
--- a/scripts/deploy-ebook-to-www
+++ b/scripts/deploy-ebook-to-www
@@ -344,7 +344,7 @@ do
 			sed --in-place --regexp-extended 's/xml:lang="([^"]+?)"/xml:lang="\1" lang="\1"/g' "${filename}"
 
 			# Add the work title to `<title>` elements in the source text.
-			sed --in-place --regexp-extended "s|<title>|<title>${workTitle} - |g" "${filename}"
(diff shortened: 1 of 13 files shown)


acabal:
I don't think we want to change this, a hyphen is standard and expected for things like page `<title>` elements, etc. 

Additionally we should never just do a wholesale change across the whole website to fix one possible issue in one place.

bentley:
I don’t agree with the first point: en dash is the typographically appropriate character (and hyphen certainly is not). It’s easy to add and there’s no technical reason to avoid it. It’s not so noticeable in `<title>`, sure, but as mentioned, I started this because I wanted to fix the instances visible in the body of the web ebooks.

I don’t agree with the second point either. This is not a wholesale change. I had to review all hyphens in the repo to make sure changing one didn’t introduce a bug somewhere else in the XSL. So naturally I replaced the other cases where en dash was more appropriate as I encountered them, and only those cases. I can split it into multiple commits if that would make the change more palatable, or drop the truly unimportant ones like the ones under contribute/.

PR #465, opened 2025-01-19
Add checks for bold/strong and illustration names/ids

vr8hub:
Add a couple of things I check and often find incorrect. I've never gotten a local web site working, but I mostly copy and pasted, so hopefully it is correct in syntax and formatting.


acabal:
Thanks!

PR #467, opened 2025-01-21
Make bold/strong regular expression catch more cases

weijia-cheng:
This regex is a little narrower than I'd like, because it can't catch elements with semantics. This revised regex will find elements with semantics too, while filtering out things like blockquotes with the use of the non-word character class.


acabal:
OK, thanks!

Issue #437, opened 2024-12-15
Keyword search too restrictive


zsau:
I [searched](https://standardebooks.org/ebooks?query=war+peace&sort=newest&view=grid&per-page=12) for the terms "war peace" (without quotes), intending to find Tolstoy's [War and Peace](https://standardebooks.org/ebooks/leo-tolstoy/war-and-peace/louise-maude_aylmer-maude). But the only result I got was [Lay Down Your Arms](https://standardebooks.org/ebooks/bertha-von-suttner/lay-down-your-arms/t-holmes). I'd likely have assumed you didn't have it and gone elsewhere, had I not known there was a translation in the public domain and that the work is such a classic.

"war and peace" (again without quotes) does find Tolstoy's work, and oddly enough doesn't list Lay Down Your Arms. Both searches behave the same with quotes, which leads me to guess that the unquoted version is being treated as though it had quotes. And I'd further guess that [this](https://github.com/standardebooks/bertha-von-suttner_lay-down-your-arms_t-holmes/blob/1c103ad8ea224a699ba677c37901baa296eb2414/src/epub/content.opf#L38-L39) is the reason Lay Down Your Arms gets returned for the query "war peace".

acabal:
When preparing the query the system puts quotes around it, so if you had searched for `war and peace` you would have found the ebook.

Not sure if removing double quotes would improve search results at all, instead of muddying them significantly. @colagrosso  what do you think?

zsau:
I don't know of any major search engine that quotes by default; it's far too restrictive. Better to return more results than less, as long as the ranking algorithm is decent.

Besides, if you don't auto-quote then users can choose to quote when they want. I'm sure there are many who don't know about quoting, yes, but the fact that quotes work in Google etc should help.

colagrosso:
Hi, sorry for the late response and that you had trouble finding the ebook you were searching for. I did look at this closely during initial development and made tweaks like d9b6e3020c3aa13ec593a8dc80cfe02455c5dba1 and 7339255d5fdc87c1155847260dfce2f908cc42e0, and I spent the morning looking for ways to make it better.

A problem is what you brought up here:

> Better to return more results than less, as long as the ranking algorithm is decent.

The search form has 4 sort orders:

* S.E. release date (new → old)  
* Author name (a → z)  
* Reading ease (easy → hard)  
* Length (short → long)

[...]

PR #470, opened 2025-01-25
Add a `Relevance` sort order and improve search

colagrosso:
I've been testing this with the queries we've discussed in #437 and a few others:

* `war peace`
* `"war and peace"`
* `sherlock`
* `tolstoy`
* `woolf`
* `poe`
* `poe poetry`
* `moby dick`
* `jane austen emma`
* `crime and punishment`                                                                                                                                                                                                                                                                  
* `to the lighthouse`                                                                                                                                                                                                                                                                     
* `great expectations`                                                                                                                                                                                                                                                                    
* `charles dickens`                                                                                                                                                                                                                                                                       
* `great expectations charles dickens`   
* `and`
* `the`
* `a`

and so on. It works well. The right books are at the top much more often than the current search. Let me know if there are other tricky queries you'd like me to try or if you'd like me to share screenshots of the above queries in action.

Here's what's in `IndexableText` right now:

1. `Title`
2. `Collections`
3. `Authors`
4. `Tags`
5. `LocSubjects`
6. `TocEntries`

(Side note that `Description` and `LongDescription` aren't in there. I [removed](https://github.com/standardebooks/web/commit/d9b6e3020c3aa13ec593a8dc80cfe02455c5dba1) them before launch because they were returning too many unrelated results.)

Here is the proposed new ranking:

```
10 * Title +
8 * Authors +
3 * Collections +
IndexableText
```

New indices for existing DBs:

```
ALTER TABLE `Ebooks` ADD COLUMN `IndexableAuthors` text NOT NULL;
ALTER TABLE `Ebooks` ADD COLUMN `IndexableCollections` text NULL;
ALTER TABLE `Ebooks` ADD FULLTEXT `indexSearchTitle` (`Title`);
ALTER TABLE `Ebooks` ADD FULLTEXT `idxSearchAuthors` (`IndexableAuthors`);
ALTER TABLE `Ebooks` ADD FULLTEXT `idxSearchCollections` (`IndexableCollections`);
```


www/ebooks/index.php
```diff
@@ -28,6 +28,14 @@
 		$view = null;
 	}
 
+	if($query != ''){
+		$queryStringParams['query'] = $query;
+		// If the user entered a query with the default sort order, change it to relevance sort.
+		if($sort == Enums\EbookSortType::Newest){
```
acabal, line 34: OK, but if the user does a search, then opts to sort by newest, won't this incorrectly force it back to relevance? We should probably add another sort type, `Default`, which would act the same as `Newest`. Or maybe there's a better way to do that.

lib/Formatter.php
```diff
@@ -31,6 +31,13 @@ public static function RemoveDiacritics(string $text): string{
 		}
 	}
 
+	/**
+	 * Remove diacritics and non-alphanumeric characters.
+	 */
+	public static function RemoveDiacriticsAndNonalphanumerics(string $text): string{
+		return trim(preg_replace('|[^a-zA-Z0-9 ]|ius', ' ', Formatter::RemoveDiacritics($text)));
```
acabal, line 38: Are you sure we want to replaced with `' '` and not `''`? Introducing spaces may unexpectedly change how matches occur.

lib/Ebook.php
```diff
@@ -1561,6 +1592,22 @@ public function Validate(): void{
 			$error->Add(new Exceptions\EbookIndexableTextRequiredException());
 		}
 
+		if(isset($this->IndexableAuthors)){
+			$this->IndexableAuthors = trim($this->IndexableAuthors ?? '');
```
acabal, line 1596: `?? ''` should not be required here because we just checked the same using `isset()` above. If you remove that does phpstan complain?

lib/Ebook.php
```diff
@@ -2424,10 +2479,22 @@ public static function GetAllByFilter(string $query = null, array $tags = [], En
...
-			$whereCondition .= ' and match(e.IndexableText) against(? in boolean mode) ';
+			// Preserve quotes in the query so the user can enter, e.g., "war and peace" for an exact match.
+			$query = trim(preg_replace('|[^a-zA-Z0-9" ]|ius', ' ', Formatter::RemoveDiacritics($query)));
+
+			$relevanceScoreField = ', (
+				match(e.Title) against (?) * ' . EBOOK_SEARCH_WEIGHT_TITLE . ' +
+				match(e.IndexableAuthors) against (?) * ' . EBOOK_SEARCH_WEIGHT_AUTHORS . ' +
+				match(e.IndexableCollections) against (?) * ' . EBOOK_SEARCH_WEIGHT_COLLECTIONS . ' +
+				match(e.IndexableText) against (?)
+			) as relevance_score ';
```
acabal, line 2490: Use CamelCase not snake_case for consistency.

lib/Ebook.php
```diff
@@ -2424,10 +2479,22 @@ public static function GetAllByFilter(string $query = null, array $tags = [], En
...
+
+			$whereCondition .= ' and match(e.IndexableText) against(?) ';
```
acabal, line 2492: Now that we match title, author, and collection separately, should we remove that data from `IndexableText`?

templates/SearchForm.php
```diff
@@ -18,13 +18,17 @@
 		</select>
 	</label>
 	<label>Keywords
-		<input type="search" name="query" value="<?= Formatter::EscapeHtml($query ?? '') ?>"/>
+		<input type="search" name="query" placeholder="Quotes allowed, e.g., &quot;war and peace&quot;" value="<?= Formatter::EscapeHtml($query ?? '') ?>"/>
```
acabal, line 21: Let's not include this placeholder, since that's standard search syntax and we're improving search anyway.

acabal (commented):
Looks pretty good! Some notes inline.

How does search perform if we use MariaDB's native relevancy sort? I believe if we do a fulltext match and don't specify a sort order then the results are by relevance. But I don't know how that behaves with multiple `match` statements.

www/ebooks/index.php
colagrosso: They can't sort by newest after they've done a search because Relevance replaces it. The first sort type is now 

* `S.E. release date` if the query is empty  
* `Relevance` if a query is present

That acts like the Default you described. I tried it out and thought it was natural. One drawback is that you can't enter a query AND sort by newest (even though you can sort by author name, reading ease, and length). That is a shame, but there's no way to tell if the user intended to do that instead of wanting to enter a new query and get results by relevance.

lib/Ebook.php
colagrosso: Done. I fixed `IndexableText` above, too. No PHPStan errors.

lib/Ebook.php

lib/Ebook.php
colagrosso: I thought about that, too, but it doesn't make things any simpler if we want to keep adding weight to title, authors, and collections. It would go like this:

1. Create a new index on all the separate columns:

```
CREATE INDEX idxSearchFull ON Ebooks (IndexableText, Title, IndexableAuthors, IndexableCollections);
```

2. Rewrite the `WHERE` clause to match against all the columns:

```
SELECT e.Title, (
    match(e.Title) against ('war peace') * 10 +
    match(e.IndexableAuthors) against ('war peace') * 8 +
    match(e.IndexableCollections) against ('war peace') * 3 +
    match(e.IndexableText) against ('war peace')) as RelevanceScore
from Ebooks e
where match(e.IndexableText, e.Title, e.IndexableAuthors, e.IndexableCollections) against('war peace')
order by RelevanceScore desc limit 12;
```

But we would still have to keep the indices against the individual columns (`Title`, `IndexableAuthors`, `IndexableCollections`, IndexableText`) so that they can match and get separate weights.

templates/SearchForm.php
colagrosso: Done.

www/ebooks/index.php
acabal: But we should be able to sort by newest after searching. I think it's reasonable to search for something general like `holmes` and want to sort by newest ebook. That's why I suggested a new `default` sorting enum, to be used internally. It would be interpreted as `newest` if there's no query, and `relevance` if there is a query.

lib/Ebook.php
acabal: What I'm saying is that right now `IndexableText` is made up of various elements including the title, collection names, authors, etc. But with this update, now we match against those things in separate columns. So do we still need to include them all in `IndexableText` along with all the other stuff in that column?

If I'm understanding things correctly then we should just be able to remove those duplicate columns from the contents of `IndexableText` without a change in results.

colagrosso: Sure, we could remove that info from `IndexableText`, but we'd have to make some other changes. Here are 3 choices. I did the first one, but the other two are just as good.

1.

Keep `IndexableText` the same.

Add these 3 indices:

```
ALTER TABLE `Ebooks` ADD FULLTEXT `indexSearchTitle` (`Title`);
ALTER TABLE `Ebooks` ADD FULLTEXT `idxSearchAuthors` (`IndexableAuthors`);
ALTER TABLE `Ebooks` ADD FULLTEXT `idxSearchCollections` (`IndexableCollections`);
```

The `WHERE` statement matches against `IndexableText`:

```
where match(e.IndexableText) against('war peace')
```

2.

Remove the extra data from `IndexableText`.

Add these 3 indices:

```
ALTER TABLE `Ebooks` ADD FULLTEXT `indexSearchTitle` (`Title`);
ALTER TABLE `Ebooks` ADD FULLTEXT `idxSearchAuthors` (`IndexableAuthors`);
ALTER TABLE `Ebooks` ADD FULLTEXT `idxSearchCollections` (`IndexableCollections`);
``` 

The `WHERE` statement matches against all 4 fields:

```
where match(e.IndexableText) against('war peace') or
   match(e.Title) against('war peace') or
   match(e.IndexableAuthors) against('war peace') or 
   match(e.IndexableCollections) against('war peace')
```

3.

Remove the extra data from `IndexableText`.

Add these 4 indices:

```
ALTER TABLE `Ebooks` ADD FULLTEXT `indexSearchTitle` (`Title`);
ALTER TABLE `Ebooks` ADD FULLTEXT `idxSearchAuthors` (`IndexableAuthors`);
ALTER TABLE `Ebooks` ADD FULLTEXT `idxSearchCollections` (`IndexableCollections`);
ALTER TABLE `Ebooks` ADD FULLTEXT `idxSearchFull` (IndexableText, Title, IndexableAuthors, IndexableCollections);
``` 

The `WHERE` statement matches against all 4 fields together :

```
where match(e.IndexableText, e.Title, e.IndexableAuthors, e.IndexableCollections) against('war peace')
```

Performance seems fine in all 3, but the explain plan for #2 looks different:

1.
```
+------+-------------+-------+----------+---------------+-----------+---------+------+------+----------------------------------------------+
| id   | select_type | table | type     | possible_keys | key       | key_len | ref  | rows | Extra                                        |
+------+-------------+-------+----------+---------------+-----------+---------+------+------+----------------------------------------------+
|    1 | SIMPLE      | e     | fulltext | idxSearch     | idxSearch | 0       |      | 1    | Using where; Using temporary; Using filesort |
+------+-------------+-------+----------+---------------+-----------+---------+------+------+----------------------------------------------+
```

2.
```
+------+-------------+-------+------+---------------+------+---------+------+------+----------------------------------------------+
| id   | select_type | table | type | possible_keys | key  | key_len | ref  | rows | Extra                                        |
+------+-------------+-------+------+---------------+------+---------+------+------+----------------------------------------------+
|    1 | SIMPLE      | e     | ALL  | NULL          | NULL | NULL    | NULL | 1401 | Using where; Using temporary; Using filesort |
+------+-------------+-------+------+---------------+------+---------+------+------+----------------------------------------------+
```

3.
```
+------+-------------+-------+----------+---------------+---------------+---------+------+------+----------------------------------------------+
| id   | select_type | table | type     | possible_keys | key           | key_len | ref  | rows | Extra                                        |
+------+-------------+-------+----------+---------------+---------------+---------+------+------+----------------------------------------------+
|    1 | SIMPLE      | e     | fulltext | idxSearchFull | idxSearchFull | 0       |      | 1    | Using where; Using temporary; Using filesort |
+------+-------------+-------+----------+---------------+---------------+---------+------+------+----------------------------------------------+
```

In all 3 cases, we need new indices on title, authors, and collections in order to put extra weight on those matches. The relevance score calculation is different from the matching in the `WHERE` statement.

www/ebooks/index.php
colagrosso: That's a good example search. I'll work on default sorting.

lib/Ebook.php
acabal: OK, I see. I think we should go with option 3, because that way we're not duplicating data in two different columns. Someone going in to hand edit the database might update `IndexableAuthors` but forget to update `IndexableText` and now the data is out of sync.

It does mean one extra index but I don't think that's a big deal in terms of performance.

So as long as the results are the same we should go with option 3.

acabal:
OK great work, I think we're ready to release! Thanks!

acabal:
It's live on the website! Great work, thanks!

Issue #419, opened 2024-11-18
Tool to handle artist duplicate/alternate names


colagrosso:
There are a few artist names in the DB that need to be corrected or added as alternate names (see [this post](https://groups.google.com/g/standardebooks/c/sQ5D27yK8JA) for an example).

Job Curtis proposed a tool to handle these cases back in Oct 2023, and in that [same thread](https://github.com/standardebooks/web/pull/289#issuecomment-1774170776) I described the manual steps that will be automated.

I've done some prototyping with a tool, and the first thing to think about is where the form will live and what the URLs will be. Here's what I propose:

Step 1: Go to the artist you want to delete. As an admin there will be a form at the bottom that reads something like "Delete this artist and reassign artwork."

Step 2: In the form, pick the canonical artist to reassign the artwork to.

Step 3: There is an optional checkbox to add an alternate name (A.K.A.) to the canonical artist.

Step 4: The form submits a `POST` request to `/artists/merge` with these parameters:

```
canonical-artist-url-name: joaquin-sorolla
alternate-artist-url-name: joaquin-sorolla-y-bastida
add-alternate-name: on
```

That is, the request parameters are the `UrlName`s of the artists. The `add-alternate-name` parameter is optional.

Here's a rough draft of what it would look like:

![Screenshot_2024-11-17_18-38-45](https://github.com/user-attachments/assets/7a411744-d453-4a6c-b08a-dbc70dbc9079)

If `/artists/merge` is successful, it will redirect to the canonical artist's page (`/artworks/joaquin-sorolla` in this case) so that the admin can see all the merged artwork together. There will be a success message at the top. If there is a problem, the admin will stay on the alternate artist, and there will be an error message at the top.

Seem reasonable?

acabal:
Sounds good. I would tweak two things:

1. Instead of showing the form on `/artists/<url-name>`, put it behind a link. This is because browsing artists is something an admin might do often, and having a prefilled form with a submit button right there is begging for an accidental click that will be very hard and annoying to undo. The form could live at `/artists/<url-name>/merge-requests/new` and we can link to that from the artist index page. That way it takes a click, then a page load, then a confirmation click.
2. Then we would POST to `/artists/<url-name>/merge-requests`. This URL maintains our object oriented (and not action oriented) URL style. The artist we are POSTing to would be the artist we are reassigning, so we don't have to pass both names as form parameters. (So the parameters would be something like `canonical-artist-id` and `add-as-alternate-name`.)

[...]

PR #480, opened 2025-03-06
Add admin tool to handle duplicate artist names

colagrosso:
Sorry for the delay.

Here's a video showing how it works:

https://github.com/user-attachments/assets/d436e269-4e12-4e7d-be76-df5495472582

It works like we described in #419

### DIFF
--- a/lib/Artist.php
+++ b/lib/Artist.php
@@ -5,6 +5,7 @@ use Safe\DateTimeImmutable;
  * @property ?int $DeathYear
  * @property ?string $UrlName
  * @property ?string $Url
+ * @property string $DeleteUrl
  * @property ?array<string> $AlternateNames
  */
 class Artist{
@@ -19,6 +20,7 @@ class Artist{
 
 	protected string $_UrlName;
 	protected string $_Url;
+	protected string $_DeleteUrl;
 	/** @var array<string> $_AlternateNames */
 	protected array $_AlternateNames;
 
@@ -42,6 +44,10 @@ class Artist{
 
 	protected function GetUrl(): string{
 		return $this->_Url ??= '/artworks/' . $this->UrlName;
+	}
+
+	protected function GetDeleteUrl(): string{
+		return $this->_DeleteUrl ??= '/artists/' . $this->UrlName . '/delete';
 	}
 
 	/**
@@ -137,6 +143,36 @@ class Artist{
 	/**
 	 * @throws Exceptions\ArtistNotFoundException
 	 */
+	public static function GetByName(?string $name): Artist{
+		if($name === null){
+			throw new Exceptions\ArtistNotFoundException();
+		}
+
+		return Db::Query('
+				SELECT *
+					from Artists
+					where Name = ?
+			', [$name], Artist::class)[0] ?? throw new Exceptions\ArtistNotFoundException('Name: ' . $name);
+	}
+
+	/**
+	 * @throws Exceptions\ArtistNotFoundException
+	 */
+	public static function GetByUrlName(?string $urlName): Artist{
+		if($urlName === null){
+			throw new Exceptions\ArtistNotFoundException();
+		}
+
+		return Db::Query('
+				SELECT *
+					from Artists
+					where UrlName = ?
+			', [$urlName], Artist::class)[0] ?? throw new Exceptions\ArtistNotFoundException('URL name: ' . $urlName);
+	}
+
+	/**
+	 * @throws Exceptions\ArtistNotFoundException
+	 */
 	public static function GetByAlternateUrlName(?string $urlName): Artist{
 		if($urlName === null){
 			throw new Exceptions\ArtistNotFoundException();
@@ -149,6 +185,28 @@ class Artist{
 					where aan.UrlName = ?
 					limit 1
 			', [$urlName], Artist::class)[0] ?? throw new Exceptions\ArtistNotFoundException();
+	}
+
+	public function AddAlternateName(string $name): void{
+		Db::Query('
+			INSERT into ArtistAlternateNames (ArtistId, Name, UrlName)
+			values (?,
+				?,
+				?)
+		', [$this->ArtistId, $name, Formatter::MakeUrlSafe($name)]);
+	}
+
+	/**
+	 * Reassigns all the artworks currently assigned to this artist to the given canoncial artist.
+	 *
+	 * @param Artist $canonicalArtist
+	 */
+	public function ReassignArtworkTo(Artist $canonicalArtist): void{
+		Db::Query('
+			UPDATE Artworks
+			set ArtistId = ?
+			where ArtistId = ?
+		', [$canonicalArtist->ArtistId, $this->ArtistId]);
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


lib/Artist.php
```diff
@@ -188,7 +246,17 @@ public static function GetOrCreate(Artist $artist): Artist{
 		}
 	}
 
+	/**
+	 * @throws Exceptions\ArtistHasArtworkException
+	 * @throws Exceptions\ArtistNotFoundException
+	 */
 	public function Delete(): void{
+		$artworkCount = count(Artwork::GetAllByArtist($this->UrlName, Enums\ArtworkFilterType::Admin, null /* submitterUserId */));
```
acabal, line 254: We should not count like this because we're potentially instantiating a huge amount of objects just to get a count. Instead do a raw `Db::QueryBool('select exists (select ...))')` which will just return true on the first record match and then stop

lib/Exceptions/ArtistHasArtworkException.php
```diff
@@ -0,0 +1,7 @@
+<?
+namespace Exceptions;
+
+class ArtistHasArtworkException extends AppException{
+	/** @var string $message */
+	protected $message = 'Artist has artwork assoicated with it. Can’t delete.';
```
acabal, line 6: Spelling error, also let's make this a complete sentence

www/artists/delete.php
```diff
@@ -0,0 +1,82 @@
...
+
+	if(!Session::$User->Benefits->CanReviewOwnArtwork){
+		throw new Exceptions\InvalidPermissionsException();
+	}
+
+	session_start();
+
+	$exception = HttpInput::SessionObject('exception', Exceptions\AppException::class);
+
+	$artist = Artist::GetByUrlName(HttpInput::Str(GET, 'artist-url-name') ?? '');
```
acabal, line 19: I don't think we need `??` here because passing null will still throw an exception, but will also avoid a DB call

www/artists/delete.php
```diff
@@ -0,0 +1,82 @@
...
+	$artworkCount = count(Artwork::GetAllByArtist($artist->UrlName, Enums\ArtworkFilterType::Admin, null /* submitterUserId */));
```
acabal, line 20: I don't think we need to calculate `$artworkCount`, it's a DB call and the number is not especially relevant to anything

www/artists/delete.php
```diff
@@ -0,0 +1,82 @@
...
+
+		<form method="<?= Enums\HttpMethod::Post->value ?>" action="/artists/<?= $artist->UrlName ?>">
+			<input type="hidden" name="_method" value="<?= Enums\HttpMethod::Delete->value ?>" />
+			<p>Are you sure you want to permanently delete <?= Formatter::EscapeHtml($artist->Name) ?>?</p>
+			<label class="icon user">
+				<span>Canonical Artist</span>
+				<span>Reassign artwork by <?= Formatter::EscapeHtml($artist->Name) ?> to this artist. Total artworks to reassign: <?= $artworkCount ?></span>
+				<datalist id="artist-names-except-this-artist">
+					<? foreach(Artist::GetAll() as $a){ ?>
+						<? if($a->ArtistId == $artist->ArtistId){ continue; } ?>
```
acabal, line 54: Instead of `continue`, test if `$a->ArtistId != $artist->ArtistId` so it's a more natural brace structure instead of a short circuit

www/artists/get.php
```diff
@@ -13,24 +16,50 @@
...
+			</ul>
+		<? } ?>
 
 		<?= Template::ImageCopyrightNotice() ?>
 
 		<?= Template::ArtworkList(artworks: $artworks) ?>
+
+		<? if($isAdminView){ ?>
+			<h2>Admin</h2>
+			<p><a href="<?= $artist->DeleteUrl  ?>">Delete artist and reassign artwork</a></p>
```
acabal, line 61: Extra space before `?>`

www/artists/post.php
```diff
@@ -0,0 +1,53 @@
...
+
+	if(!Session::$User->Benefits->CanReviewOwnArtwork){
+		throw new Exceptions\InvalidPermissionsException();
+	}
+
+	// DELETE an `Artist`.
+	if($httpMethod == Enums\HttpMethod::Delete){
+		$artistToDelete = Artist::GetByUrlName(HttpInput::Str(GET, 'artist-url-name') ?? '');
+		$exceptionRedirectUrl = $artistToDelete->DeleteUrl;
+		$artworkReassignedCount = count(Artwork::GetAllByArtist($artistToDelete->UrlName, Enums\ArtworkFilterType::Admin, null /* submitterUserId */));
```
acabal, line 19: Not necessary to calculate

acabal (changes requested):
Great work Mike, thanks! I made some comments inline. In general I don't think we need to calculate the number of artworks changed, it's a DB call and that info isn't especially relevant.

colagrosso:
Thanks for the quick review. I pushed the fixes as a separate commit so that they are easier to review. You can squash and merge them when we're done.

lib/Artist.php
```diff
@@ -151,6 +187,28 @@ public static function GetByAlternateUrlName(?string $urlName): Artist{
 			', [$urlName], Artist::class)[0] ?? throw new Exceptions\ArtistNotFoundException();
 	}
 
+	public function AddAlternateName(string $name): void{
+		Db::Query('
```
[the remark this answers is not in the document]
colagrosso: I added logic to catch the duplicate key exception and show a useful error message to the user.

lib/Artist.php
```diff
@@ -134,6 +140,36 @@ public static function GetAll(): array{
...
+	public static function GetByName(?string $name): Artist{
+		if($name === null){
+			throw new Exceptions\ArtistNotFoundException();
+		}
+
+		return Db::Query('
+				SELECT *
+					from Artists
+					where Name = ?
+			', [$name], Artist::class)[0] ?? throw new Exceptions\ArtistNotFoundException('Name: ' . $name);
```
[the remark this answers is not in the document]
colagrosso: Done

lib/Artist.php
```diff
@@ -134,6 +140,36 @@ public static function GetAll(): array{
...
+	public static function GetByUrlName(?string $urlName): Artist{
+		if($urlName === null){
+			throw new Exceptions\ArtistNotFoundException();
+		}
+
+		return Db::Query('
+				SELECT *
+					from Artists
+					where UrlName = ?
+			', [$urlName], Artist::class)[0] ?? throw new Exceptions\ArtistNotFoundException('URL name: ' . $urlName);
```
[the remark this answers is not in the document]
colagrosso: Done

lib/Artist.php
colagrosso: Done

lib/Exceptions/ArtistHasArtworkException.php
colagrosso: I took another whack at rewriting the error message.

www/artists/delete.php
colagrosso: Done

www/artists/delete.php
colagrosso: Done

www/artists/delete.php
colagrosso: Done

www/artists/get.php
colagrosso: Done

www/artists/post.php
colagrosso: Done

acabal:
Great work, thanks!

PR #481, opened 2025-03-07
Clarify the files excluded from language check

vr8hub:
I modified the wording and added a couple of files so they would have the complete list. I didn't bother with including `uncopyright`, since we already check that it is exactly the same as the template. Nothing else qualifies as boilerplate, so I just said "and title page."


www/contribute/producing-an-ebook-step-by-step.php
```diff
@@ -191,7 +191,7 @@
...
-				<p>Note that we <em>don’t</em> change the language for the metadata or front/back matter files, like <code class="path">content.opf</code>, <code class="path">titlepage.xhtml</code>, or <code class="path">colophon.xhtml</code>. Those must always be in American spelling, so they’ll always have the en-US language tag.</p>
+				<p>Note that we <em>don’t</em> change the language for the metadata or title page files, like <code class="path">content.opf</code>, <code class="path">colophon.xhtml</code>, <code class="path">imprint.xhtml</code>, <code class="path">toc.xhtml</code>, or <code class="path">titlepage.xhtml</code>. Those must always be in American spelling, so they’ll always have the en-US language tag.</p>
```
acabal, line 194: Let's word this as `or boilerplate files` because the ToC/imprint/etc. are not really "title page files". Let's also cut  the explicit reference to content.opf because that's the only metadata file with lang anyway.

vr8hub: Of course they're not "title page" files, only titlepage is, but I consider them metadata, or at least metadata-adjacent. Certainly more metadata than boilerplate. Uncopyright is boilerplate, these aren't that. Nevertheless, changes made.

acabal:
Thanks!

PR #486, opened 2025-03-21
Remove non-alphanumerics (except quotes) from search query

### LINKED ISSUE #484: Searching for compound words doesn't appear to work
I just tried searching for `Haycraft-Queen` on the [Books page](https://standardebooks.org/ebooks), and it returned a bunch of books that had "Queen" somewhere, but none of the Haycraft-Queen series books.

I then tried it with quotes around it, `"Haycraft-Queen"`, and it returned no books.

I then tried just `Haycraft`, and that didn't return anything, either.

The combination of those don't make sense to me—it seems that it's splitting the two words and searching for them individually, but it's not finding either of them in the "Haycraft-Queen" series name?

I then tried `Haycraft-Queen Cornerstones`, and it did return the Haycraft-Queen items (because of `Cornerstones` presumably), but it also returned anything with just `queen` in it (and, I'm assuming, anything with just `cornerstones` in it, but there was 18 pages and I didn't go them them all :) ). Putting quotes around the whole thing, `"Haycraft-Queen Cornerstones"`, again returned nothing.

Pulling up one of the series' books and clicking on the collection name does return all the books in the collection, but surely we should be able to search for the collection name to find them?

I would think the interface should search for a compound word as a whole before it searches for it's component parts?

Also, shouldn't we be able to put quotes around something to have it search for the entire term rather than individual pieces of it? In this case I don't want the "queen"s, I just want the "Haycraft-Queens."

colagrosso:
That is, don't replace non-alphanumerics with a space. This matches the behavior of Formatter::RemoveDiacriticsAndNonalphanumerics(), which would be used here except that function would also remove quotes.  We actually discussed not introducing spaces previously, but I made a mistake and didn't apply the same change to the user's search query: https://github.com/standardebooks/web/pull/470#discussion_r1929591492

This change fixes queries with compound words like:

`Haycraft-Queen`

and also fixes queries for authors with apostrophes:

`O'Neill`

No changes to existing DBs are necessary because they already have terms like `haycraftqueen` and `oneill` stored in `IndexableCollections` and `IndexableAuthors`.

Fixes #484

### DIFF
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -2333,7 +2333,7 @@ final class Ebook{
 
 		if($query !== null && $query != ''){
 			// Preserve quotes in the query so the user can enter, e.g., "war and peace" for an exact match.
-			$query = trim(preg_replace('|[^a-zA-Z0-9" ]|ius', ' ', Formatter::RemoveDiacritics($query)));
+			$query = trim(preg_replace('|[^a-zA-Z0-9" ]|ius', '', Formatter::RemoveDiacritics($query)));
 
 			$whereCondition .= ' and match(e.IndexableText, e.Title, e.IndexableAuthors, e.IndexableCollections) against(?) ';
 			$params[] = $query;
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -1706,6 +1706,7 @@ final class Ebook{
 			}
 		}
 
+		$this->IndexableText = str_replace('-', ' ', $this->IndexableText);
 		$this->IndexableText = Formatter::RemoveDiacriticsAndNonalphanumerics($this->IndexableText);
 
 		if($this->IndexableText == ''){
@@ -1719,6 +1720,7 @@ final class Ebook{
 			$this->IndexableAuthors .= ' ' . $author->Name;
 		}
 
+		$this->IndexableAuthors = str_replace('-', ' ', $this->IndexableAuthors);
 		$this->IndexableAuthors = Formatter::RemoveDiacriticsAndNonalphanumerics($this->IndexableAuthors);
 
 		// Initialize `IndexableCollections`.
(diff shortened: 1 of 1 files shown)


colagrosso:
Based on the discussion in #484, I added a special case for replacing hyphens with spaces. Let me know how it looks to you and if you want me to move it into a `Formatter` function (or combined with `Formatter::RemoveDiacriticsAndNonalphanumerics()`).

With this change users can search for these terms:

`beta` to match `Alpha-Beta`
`queen` to match `Haycraft-Queen`

These searches also work as expected:

`Alpha-Beta`
`Alpha`
`Haycraft-Queen`
`Haycraft`

I don't think these queries should work, and they do not:

`AlphaBeta`
`HaycraftQueen`

This commit changes `IndexableText`, `IndexableAuthors`, and `IndexableCollections`, so existing DBs need an update. This will update all published books:

```
cd /standardebooks.org/ebooks
for BOOK in $(find /standardebooks.org/ebooks -maxdepth 1 -type d)
do
  tsp nice /standardebooks.org/web/scripts/deploy-ebook-to-www --verbose --no-build --no-images --no-recompose --no-epubcheck --no-feeds --no-bulk-downloads "$BOOK"
done
```

And this PHP code will update placeholders:

```
<?
require_once('/standardebooks.org/web/lib/Core.php');

$ebooks = Ebook::GetAll();

foreach($ebooks as $ebook){
        if($ebook->IsPlaceholder()){
                print('Saving ' . $ebook->Identifier . "\n");

                // Need to force `Ebook::GetAllContributors()` to be called before `Ebook::Save()`. Otherwise, authors and translators will be deleted.
                $ebook->Authors;

                $ebook->Save();
        }
}
```

acabal:
OK great. This looks good, I'll merge it in and start the updates now. Thanks!

PR #494, opened 2025-04-24
Add Ebook::DownloadUrl for web and feed downloads

colagrosso:


colagrosso:
Right on, thanks for the quick review!

I renamed the method to `GetDownloadUrl()`. I also removed `EbookDownloadSource::Unknown` and just used `null` to mean unknown. It made the code shorter and more readable. During development, I thought I should try to be explicit about when a source was provided or accidentally left off, but it didn't end up that way.

acabal:
Thanks!

PR #495, opened 2025-04-28
Update producing-an-ebook-step-by-step.php

dajare:
Add a 'tip' in the Semantics phase to point new producers to using local.css for extra formatting rules.


acabal:
I don't want to put too much stuff into the step by step because then it gets crowded and unfocused. Where to put CSS is stated in the manual and also obvious from even the most cursory inspection of an existing ebook; and in the past ~10 years I can count the number of people who asked about it on one hand.

dajare:
👍 I had a feeling that might be the case! Understood.

PR #499, opened 2025-05-20
Identifiers no longer prefixed with `url:`

brendanny:
Addresses https://github.com/standardebooks/tools/issues/831

Now that Identifier and FullUrl are identical we can remove one of them from Ebook.php, but that might be confusing for the different consumers. (e.g. ones that are web-first (rss/web) vs ones that are epub-first (database/opds?)).

I think a database command is required to remove `url:` from `Identifier` column.

### DIFF
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -1147,7 +1147,7 @@ final class Artwork{
 		$ebookUrl = HttpInput::Str(POST, 'artwork-ebook-url');
 		if(isset($ebookUrl)){
 			try{
-				$ebook = Ebook::GetByIdentifier('url:' . $ebookUrl);
+				$ebook = Ebook::GetByIdentifier($ebookUrl);
 				$this->EbookId = $ebook->EbookId;
 			}
 			catch(Exceptions\EbookNotFoundException){
--- a/lib/Constants.php
+++ b/lib/Constants.php
@@ -35,7 +35,7 @@ const MANUAL_PATH =			WEB_ROOT . '/manual';
 const EBOOKS_DIST_PATH =		WEB_ROOT . '/ebooks/';
 const COVER_ART_UPLOAD_PATH =		'/images/cover-uploads/';
 
-const EBOOKS_IDENTIFIER_ROOT =		'url:https://standardebooks.org';
+const EBOOKS_IDENTIFIER_ROOT =		'https://standardebooks.org';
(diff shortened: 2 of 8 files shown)


acabal:
We should keep both properties because the identifier may change again in the future. The concept of the book's identifier and its URL are separate.

brendanny:
That's what I was hung up on, too, but couldn't think of any good reason they would not be identical. By the spec, the identifier **is** the url. If the url changes the identifier changes (and vice versa). Conceptually, they seem different, but it seems an artifact from having them be different in the past.

acabal:
Just like we changed it now, we might change it again in the future. Therefore we should keep them as separate oncepts.

acabal:
@colagrosso Mike would you be interested in reviewing this for correctness?

See the issue in the top post.

colagrosso:
Yes! I will review it this morning.

colagrosso (commented):
Thanks, Brendan, for working on a large change across multiple repos! The changes to the website look good to me, but wait for Alex to approve.

To test it locally, I had to update the `Identifier` strings in my database with this SQL:

```sql
UPDATE Ebooks
SET Identifier = REPLACE(Identifier, 'url:', '')
WHERE Identifier LIKE 'url:%';
```

I ran this `SELECT` statement first to double-check that the `REPLACE` would work as expected:

```sql
SELECT EbookId, Identifier AS OriginalIdentifier, REPLACE(Identifier, 'url:', '') AS NewIdentifier
FROM Ebooks
WHERE Identifier LIKE 'url:%'
LIMIT 20;
```

After that, the website worked as expected. I tested:

1. Browsing, searching, and viewing ebooks
1. Adding new artwork
1. Editing the ebook URL in the artwork form

Then I tested adding/updating ebooks:

1. Take a published ebook, update its `<dc:identifier>` to remove `url:` and rerun `scripts/deploy-ebook-to-www` to check the `UPDATE` statement.
1. Download the repo for a new ebook, update its `<dc:identifier>` to remove `url:` and run `scripts/deploy-ebook-to-www` to check the `INSERT` statement.

I checked that there weren't any code lingering references to `url:` strings in the repo. There aren't.

Things I didn't test yet: 
1. Since the corpus on Github hasn't changed yet, I didn't test `scripts/sync-ebooks`, but the change looks good.

Since we're changing these lines already, I considered whether properties like `GetUrlSafeIdentifier()` and `GetAuthorsUrl()` should use `EBOOKS_IDENTIFIER_PREFIX` instead of `https://standardebooks.org/ebooks/`, but I decided these are fine as-is. Maybe we will change them in a future clean up.

PR #508, opened 2025-07-17
Limit ebook downloads per IP address

colagrosso:
Feedback welcome on the thresholds and time periods in addition to anything else.

I decided not to add an index on `IpAddress` in `EbookDownloads` because the query currently used the index on `Created`, which seems fine:

```sql
MariaDB [se]> explain SELECT count(*) from EbookDownloads where IpAddress = '::ffff:127.0.0.1' and Created >= '2025-07-16 00:21:47';
+------+-------------+----------------+-------+---------------+-------------+---------+------+------+------------------------------------+
| id   | select_type | table          | type  | possible_keys | key         | key_len | ref  | rows | Extra                              |
+------+-------------+----------------+-------+---------------+-------------+---------+------+------+------------------------------------+
|    1 | SIMPLE      | EbookDownloads | range | idx_created   | idx_created | 4       | NULL | 61   | Using index condition; Using where |
+------+-------------+----------------+-------+---------------+-------------+---------+------+------+------------------------------------+
```


www/ebooks/download.php
```diff
@@ -40,11 +45,15 @@
 	/** @var string|null $userAgent */
 	$userAgent = $_SERVER['HTTP_USER_AGENT'] ?? null;
 
-	try{
-		$ebook->AddDownload($ipAddress, $userAgent);
+	// Check for excessive downloads.
```
acabal, line 48: If the user is logged in,  or if we're from a `source`, then we can skip rate limiting downloads, on the assumption that patrons are going to be polite.

acabal (changes requested):
Great! A few notes.

www/ebooks/download.php
```diff
@@ -56,6 +65,13 @@
 			throw new Exceptions\InvalidFileException();
 		}
 
+		try{
```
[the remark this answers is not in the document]
colagrosso: I meant to leave a comment about this when I sent the PR: I think we are currently recording downloads twice when the user sees the download page. (We were on my local machine.) The download page records the download and it contains a refresh tag:

```html
<meta http-equiv="refresh" content="0; url=<path/to/epub>?source=download" />
```

that also records the download.

I'm sorry about that mistake. I used this SQL to see if I could find `EbookDownloads` with the same `EbookId, IpAddress, UserAgent) within 2 seconds of each other:


```sql
WITH LaggedDownloads AS (
    SELECT
        EbookId,
        Created,
        IpAddress,
        UserAgent,
        LAG(Created, 1) OVER (
            PARTITION BY
                EbookId,
                IpAddress,
                UserAgent
            ORDER BY
                Created ASC
        ) AS PreviousCreated
    FROM
        EbookDownloads
)
SELECT
    *
FROM
    LaggedDownloads
WHERE
    PreviousCreated IS NOT NULL
    AND TIMESTAMPDIFF(SECOND, PreviousCreated, Created) <= 2;
```

You can use it to look at your own DB.

www/ebooks/download.php
colagrosso: I added a check if the user is logged in, but since all file downloads come from `source=feed` or `source=download`, I didn't add a check for `source`. (I also thought it would be each for someone writing a script to add.)

acabal:
OK, no problem, we can just wipe the data for now, it's only been about a month.

The next thing to think about is how to apply this to the bot count in download summaries. If someone hits this rate limit they're probably a bot and should be counted as such for purposes of the summaries. Right now we only check the user-agent. So I think we might need an additional table to track which IPs get rate limited, to be used when regenerating download summaries and then cleared afterwards.

colagrosso:
Right on, thanks for understanding. I'll work on the new summary logic for download summaries next.

acabal:
I would also note that [Project Gutenberg does the exact same thing](https://www.gutenberg.org/policy/terms_of_use.html#:~:text=Make%20no%20more,to%20%60hack%20around%C2%B4.).

Issue #510, opened 2025-07-22
Page crash at `blog/public-domain-day-2025` in chromium browsers


brendanny:
Visiting https://standardebooks.org/blog/public-domain-day-2025 crashes the tab with `Error code: 11`.

Crashes in stable releases of Google Chrome, Brave, and Chromium (v138) and Google Chrome Canary (v140). Does not crash and loads fine in Firefox or Safari.

All other blog posts at `blog/` work fine.

colagrosso:
I'll try not to spam this issue, but if you're also looking at it and it saves you time, I can fix the issue by removing the images here:

https://github.com/standardebooks/web/blob/master/templates/RealisticEbook.php

So your comment above appears correct. I'll keep looking.

colagrosso:
At the risk of giving away how little I understand CSS, removing [this](https://github.com/standardebooks/web/blob/af205f39ea9e59ebf883087c41dc2ec35126456b/www/css/public-domain-day.css#L5-L8) second rule in `public-domain-day.css` breaks the formatting but also prevents the `SIGSEGV`. Of course, there are subsequent rules that probably depend on that one.

```
.public-domain-day li{
	display: flex;
	margin-top: 6rem;
}```

**Comment by robinwhittleton** [comment]
Working on a minimal test case now, at least that’ll give us some idea of what’s going wrong, and something to submit to Chromium.

**Comment by robinwhittleton** [comment]
Doesn’t actually help us with a fix, but this is as small as I can get that reliably crashes Chromium. Filed at https://issues.chromium.org/issues/433533357.

```
<!DOCTYPE html>
<html>

<head>
    <meta charset="utf-8" />
    <title>Test</title>
    <style>
    body {
        display: flex;
        text-wrap: pretty;
    }

    figure::before {
        content: "";
        position: absolute;
    }

    img {
        display: block;
    }
    </style>
</head>

<body>
    <div>
        <a>
            <figure>
                <img />
            </figure>
        </a>
    </div>
</body>

</html>
```

robinwhittleton:
OK, I’d suggest removing `text-wrap: pretty` from core.css: https://github.com/standardebooks/web/blob/af205f39ea9e59ebf883087c41dc2ec35126456b/www/css/core.css#L185. That stops the crashing, for a minimal decrease in rendering quality.

acabal:
[...]

PR #511, opened 2025-07-22
Remove the <div> around ebook images

colagrosso:
I thought maybe simplifying the DOM would help, and removing this `<div>` seems to prevent the crash reported in #510.

It's not a satisfying solution, so I can look for a different approach if you prefer.


acabal:
Looks like this was caused by a Chromium error involving the `text-wrap: pretty` CSS rule.

PR #514, opened 2025-08-13
Correct error in blog post

erinendrei:
To make sense the word "ethics" needs to refer to the subject, not to Spinoza's book _Ethics_.


acabal:
Thanks!

PR #515, opened 2025-08-14
Update sync-ebooks

vr8hub:
With the change to the URL scheme, i.e. eliminating the leading "url:", `sync-ebooks` needs a corresponding change, as it looks for URL's in the metadata at one point.

Also, when I originally added the comment to the code for the renaming of the repo to the identifier, I didn't know what was happening or why, and said that. I know both now, so this also updates the comment accordingly.

I also have two local changes that I want to see if you'd be interested in.
The first one adds a `--new-only` option, like the current `--update-only`, and does what you would expect. I often want to get the new books, but don't want to take the time at that moment to update the entire corpus. (This happens more often now that the corpus has gotten so large.)

The second one adds a `--start-letter` option, again as a result of the size of the corpus. If something happens during the update that causes it to die, e.g. forgotten local changes that shouldn't be there, a network glitch, whatever, having to restart from the beginning is a pain and waste of time. This lets me specify a starting letter, so I can re-start at P, or wherever, and not have to wait through the hundreds of books prior being processed again to no effect.

### DIFF
--- a/scripts/sync-ebooks
+++ b/scripts/sync-ebooks
@@ -245,9 +245,9 @@ printf "%s\n" "${repoUrls}" | while IFS= read -r repoUrl; do
 
 	# get the last segment of the dc:identifier from the metadata
 	properName="$(git -C "${repoName}" show HEAD:src/epub/content.opf |
-		grep -oE "<dc:identifier id=\"uid\">url:https://standardebooks.org/ebooks/[^<]+</dc:identifier>" |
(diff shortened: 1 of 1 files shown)


acabal:
I'm not sure how this PR is possible, `url:` has already been removed from `sync-ebooks`, see https://github.com/standardebooks/web/blob/master/scripts/sync-ebooks

What am I missing here?

PR #522, opened 2025-08-29
Add an `$updateDownloads` parameter to `CreateOrUpdate()`

### LINKED ISSUE #521: Modifying books seems to reset download counter
If you sort the books by popularity and look at the least popular books there are some surprising titles, such as _Right Ho, Jeeves_, and _Vanity Fair_. It appears that all the least popular books have been modified recently; I suspect that this must be resetting the download counter somehow.

colagrosso:
We don't want `Ebook::FromFilesystem()` to reset the downloads to 0.

Fixes #521

### DIFF
--- a/lib/Ebook.php
+++ b/lib/Ebook.php
@@ -1456,14 +1456,16 @@ final class Ebook{
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


acabal:
Looks good, thanks! I'll just delete the existing download data and we'll start fresh, considering the other issue we fixed the other day.

PR #525, opened 2025-08-30
Shorten how `COALESCE` fields are saved

colagrosso:
Sorry for the churn. This PR doesn't change any functionality. I just thought of a shorter way we can handle `DownloadsPast30Days` and `DownloadsTotal`.

This is better because we will probably add more properties to `Ebook` in the future, and this code makes it clearer how to add more properties because it keeps the same pattern as the other objects.


acabal:
Thanks!

PR #531, opened 2025-09-09
Sort by `GetSortedName()` in `GetAll()`

colagrosso:
I was working on bulk downloads, and I noticed that this page:

https://standardebooks.org/bulk-downloads/collections

is sorted by `GetSortedName()`, but this page isn't:

https://standardebooks.org/collections

### DIFF
--- a/lib/Collection.php
+++ b/lib/Collection.php
@@ -82,13 +82,26 @@ class Collection{
 
 	/**
 	 * @return array<Collection>
+	 *
+	 * @throws Exceptions\AppException
 	 */
 	public static function GetAll(): array{
-		return Db::Query('
+		$collections = Db::Query('
 					SELECT *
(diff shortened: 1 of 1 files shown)


acabal:
Good catch, thanks!

acabal:
Although it just occurred to me, why don't we skip the collator and do `select * from Collections order by regexp_replace(Name, '^(a|the|an|and)\\s', '') asc`? 

IIRC the collator is only used when sorting outside of SQL, for example in bulk downloads where there is no DB involved, and PHP doesn't natively know how to sort letters with diacritics. But I think SQL does correctly sort them, the collation is `utf8mb4_general_ci` which should be basically fine.

colagrosso:
Sure, good call. I'll send another PR.

PR #535, opened 2025-09-21
Add set/series reminder for metadata, how-to-review-an-ebook-producti…

dajare:
…on...
This seems too often to fly under the radar, and I'd find it helpful to have an explicit step/reminder.


acabal:
Thanks!

PR #536, opened 2025-09-24
add www to fracturedatlas.org url

gmoothart:
www is needed to avoid an ssl error in the browser.

arguably fracturedatlas should have a cert that covers this, but it appears that it doesn't.


acabal:
Good catch, thanks!

PR #555, opened 2026-02-05
Filter artwork by date range

### LINKED ISSUE #502: Filter artwork by date range?
Would it be possible/practical to add a filter to the artwork search that would allow us to only see artwork with the year completed being within a specified date range? E.g. if we're looking for artwork for an 1875 book, we only want to see artwork created within a range of years around that date.

Having a list of date ranges—1700's/ 1800's/etc. or 1700–1750/1751–1800/etc.— to choose from would be OK, but optimally we could just specify a start/end year.

colagrosso:
Fixes #502

Feedback welcome, especially on the styling. Here's what the artwork form looks like, now 2 rows:
Screenshot_2026-02-04_18-58-34

At 760px, it goes to 3 rows:
Screenshot_2026-02-04_18-59-49

### DIFF
--- a/lib/Artwork.php
+++ b/lib/Artwork.php
@@ -1000,48 +1000,58 @@ final class Artwork{
 	/**
 	* @return array{artworks: array<Artwork>, artworksCount: int}
 	*/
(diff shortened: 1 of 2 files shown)


colagrosso:
Thanks for the quick review, and thanks a ton for the followup commit 197143ade9ff1af2647e4b5421ea85f47ff22784. It looks 100% better.

PR #569, opened 2026-05-14
Correct elision description from semicolons to ellipsis

Robotathand:
From looking at the example, I believe this should say ellipsis instead of semicolon.


acabal:
Thanks!

PR #571, opened 2026-06-03
Provide command to search for italics in step-by-step guide

weijia-cheng:
I copied the command from the review guide, since producers (especially first-timers) would benefit from having an example of how to search for italics.


acabal:
Thanks!

PR #575, opened 2026-07-13
Fix alphabet (duplicate letter o)

MarcelWaldvogel:
The alphabet for the `--start-letter` option contained a superfluous `o`. As far as I followed the start-letter code path, it does not impact the correctness of the code. Still, it might hurt sometime in the future.


acabal:
Thanks!

